PR #9, opened 2019-07-10
put/get

davidmarkclements:
* salt (just for mutable - like bt dht?) [done]
* tests [will do in separate PR]

this is ready to merge on approval


RangerMauve:
This'll be great to have, thanks! :D

stores.js
```diff
@@ -0,0 +1,179 @@
...
+      } // silently ignore bad values
+    })
+
+    finished(queryStream, (err) => {
+      if (err) {
+        cb(err)
+        return
+      }
+      if (found === false) {
+        cb(Error('Not Found'))
```
mafintosh: i prefer cb(null, null) to indicate it does not exist

stores.js
```diff
@@ -0,0 +1,179 @@
...
+    })
+  },
+  command: {
+    update ({ target, value }, cb) {
+      if (value == null) {
+        cb(null)
+        return
+      }
+      const key = target.toString('hex')
+      store.set(key, value)
```
mafintosh: untrusted input so you need to verify that the target === hash(value)

stores.js
```diff
@@ -0,0 +1,179 @@
...
+    if (typeof opts !== 'object') throw Error('Options are required')
+    const { salt, seq } = opts
+    if (typeof seq !== 'number') throw Error('seq is a required option')
+    if (salt && !Buffer.isBuffer(key)) throw Error('salt must be a buffer')
+    const queryStream = this.query('mutable-store', key)
+    let found = false
+    queryStream.on('data', (result) => {
+      if (result.value === null) return
+      const { key, value, sig, seq: _seq, salt } = result.value
+      if (_seq === seq && verify(sig, value, key)) {
```
mafintosh: _seq is a min clock so _seq <= seq

stores.js
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+})
+
+const mutable = (store) => ({
+  keypair () {
+    const pk = Buffer.alloc(pkSize)
+    const sk = Buffer.alloc(skSize)
+    createKeypair(pk, sk)
+    return { pk, sk }
```
mafintosh: name these publicKey and secretKey

stores.js
```diff
@@ -0,0 +1,179 @@
...
+    if (typeof seq !== 'number') throw Error('seq is a required option')
+    if (salt && !Buffer.isBuffer(key)) throw Error('salt must be a buffer')
+    const queryStream = this.query('mutable-store', key)
+    let found = false
+    queryStream.on('data', (result) => {
+      if (result.value === null) return
+      const { key, value, sig, seq: _seq, salt } = result.value
+      if (_seq === seq && verify(sig, value, key)) {
+        found = true
+        cb(null, value, { key, sig, seq, salt })
```
mafintosh: We prob need an option to continue scanning to see if a newer one comes along (can be v2 feature)

stores.js
```diff
@@ -0,0 +1,179 @@
...
+    if (value.length > PUT_VALUE_MAX_SIZE) { throw Error(`Value size must be <= ${PUT_VALUE_MAX_SIZE}`) }
+    if (typeof opts !== 'object') throw Error('Options are required')
+    if (typeof cb !== 'function') throw Error('Callback is required')
+    if (value.length > PUT_VALUE_MAX_SIZE) { throw Error(`Value size must be <= ${PUT_VALUE_MAX_SIZE}`) }
+    const { seq, salt, keypair } = opts
+    if (!keypair) throw Error('keypair is required')
+    const { sk, pk } = keypair
+    if (!Buffer.isBuffer(sk)) throw Error('keypair.sk (secret key buffer) is required')
+    if (!Buffer.isBuffer(pk)) throw Error('keypair.pk (public key buffer) is required')
+    if (typeof seq !== 'number') throw Error('seq is a required option')
```
mafintosh: seq should be optional (=0 default)

stores.js
```diff
@@ -0,0 +1,179 @@
...
+  command: {
+    valueEncoding: mutableEncoding,
+    update (result, cb) {
+      if (result.value == null) {
+        cb(null)
+        return
+      }
+      const key = result.target
+      const { value, sig, seq } = result.value
+      store.set(key.toString('hex'), { key, value, sig, seq })
```
mafintosh: untrusted input, verify the signature and that the seq is >= your local seq if you have an earlier value

stores.js
```diff
@@ -0,0 +1,179 @@
...
+      cb(null)
+    },
+    query ({ target }, cb) {
+      const key = target.toString('hex')
+      const result = store.get(key)
```
mafintosh: only reply back if your local seq is >= the one sent over the wire

stores.js
```diff
@@ -0,0 +1,179 @@
...
+    if (salt) {
+      if (!Buffer.isBuffer(salt)) throw Error('salt must be a buffer')
+      if (salt.length >= 16 && salt.length <= 64) { throw Error('salt length must be between 16 and 64 bytes (inclusive)') }
+    }
+    const sig = Buffer.alloc(signSize)
+    sign(sig, value, sk)
+    const key = pk
+    const info = { key, value, sig, seq }
+
+    this.update('mutable-store', key, info, (err) => {
```
mafintosh: nit: instead of these methods being implicitly bound to this === dhtNode make this a class that accepts a dht instance in the constructor. easier to read (if you ask me) and removes the need for the bind in index.js

stores.js
```diff
@@ -0,0 +1,179 @@
...
+    if (o == null) return o
+    return MutableStore.encode(o)
+  },
+  decode (o) {
+    if (o == null) return o
+    return MutableStore.decode(o)
+  }
+}
+
+const PUT_VALUE_MAX_SIZE = 1000
```
mafintosh: Add a comment that PUT_VALUE_MAX_SIZE + packet overhead (ie the key etc) should be < the network MTU which is normally 1400 bytes

stores.js
davidmarkclements: so for that case could we just allow no callback to be passed, and then return a stream? (which would be a transform stream that query stream is connected to?

mafintosh: great idea, stream should contain every hit >= seq

stores.js
davidmarkclements: I've made it optional for both get and put (with 0 default)

stores.js
davidmarkclements: or in this case when `_seq <= seq` right?

mafintosh: yes

stores.js
```diff
@@ -0,0 +1,260 @@
...
+  }
+}
+class MutableStore {
+  constructor (dht, store) {
+    this.dht = dht
+    this.store = store
+  }
+  salt (size = 16) {
+    if (size < 16 && size > 64) {
+      console.log(size)
```
mafintosh: stray log

stores.js
```diff
@@ -0,0 +1,260 @@
...
+      }
+    }
+  }
+}
+class MutableStore {
+  constructor (dht, store) {
+    this.dht = dht
+    this.store = store
+  }
+  salt (size = 16) {
```
mafintosh: special reason for salt to 16? otherwise it's safer to default it to 32 bytes unless you know what you are doing

stores.js
```diff
@@ -0,0 +1,260 @@
...
+    const secretKey = Buffer.alloc(skSize)
+    createKeypair(publicKey, secretKey)
+    return { publicKey, secretKey }
+  }
+  get (key, opts, cb) {
+    if (typeof opts !== 'object') throw Error('Options are required')
+    const { dht } = this
+    const { salt, seq = 0 } = opts
+    if (typeof seq !== 'number') throw Error('seq should be a number')
+    if (salt && !Buffer.isBuffer(key)) throw Error('salt must be a buffer')
```
mafintosh: nit: let's move all the throws to asserts where applicable

stores.js
```diff
@@ -0,0 +1,260 @@
...
+          cb(null)
+          return
+        }
+        const publicKey = input.target
+        const { value, salt, sig, seq } = input.value
+        const key = salt
+          ? publicKey.toString('hex') + salt.toString('hex')
+          : publicKey.toString('hex')
+        const local = store.get(key)
+        const msg = salt ? Buffer.concat([value, salt]) : value
```
mafintosh: this has security issues (ie value='a', salt='a') === (value='aa', salt='').
better encoding: salt-length-in-one-byte + salt + value OR use a hash(value, key=salt)

stores.js
```diff
@@ -0,0 +1,260 @@
...
+        const verified = verify(sig, msg, publicKey) &&
+          (local ? seq >= local.seq : true)
+
```
mafintosh: okay small detail i forgot to mention ... if seq === local.seq AND value === local.value (and the sig is correct obvs) then simply refresh the LRU (ie not an error). this happens all the time in practice as nodes will republish data to make sure it stays alive.

mafintosh:
almost there :) we prob need some tests also

stores.js
davidmarkclements: only reason was to reduce payload sizes - will change

stores.js
davidmarkclements: I think this behaviour is already there..

if seq === local.seq will pass seq >= local.seq - so if the sig verifies and seq === local.seq, then `verified` will be true, so `store.set(key, { value, salt, sig, seq })` will execute which will refresh the LRU by re-setting the value. There's no need to check that values equal for this case because the behaviour would be the same whether it's changed or not

davidmarkclements: clarified offline, now I understand

stores.js
```diff
@@ -0,0 +1,301 @@
...
+      assert(
+        salt.length >= 16 && salt.length <= 64,
+        'salt size must be between 16 and 64 bytes (inclusive)'
+      )
+    }
+    const queryStream = dht.query('mutable-store', key, { salt, seq })
+    let found = false
+    const userSeq = seq
+    const streamMode = typeof cb !== 'function'
+    const responseStream = Transform({
```
mafintosh: I just remembered we have a map option in dht-rpc for exactly this purpose (ie value transform)!

We should use that instead of the transform, query(..., { map(data) { return newData } })

mafintosh: See here for an example, https://github.com/hyperswarm/dht/blob/master/index.js#L48

Getting the dht node out as well is super helpful like we do in the peer lookup :)

stores.js
```diff
@@ -0,0 +1,273 @@
...
+        }
+        if (found === false) cb(null, null)
+      })
+    }
+    return queryStream
+  }
+  put (value, opts, cb) {
+    assert(Buffer.isBuffer(value), 'Value must be a buffer')
+    assert(typeof opts === 'object', 'Options are required')
+    assert(typeof cb === 'function', 'Callback is required')
```
mafintosh: nit, cb should be optional (default to a noop)

stores.js
```diff
@@ -0,0 +1,273 @@
...
+        const msg = salt
+          ? Buffer.concat([Buffer.from([salt.length]), salt, value])
+          : value
+
+        if (local && local.seq === seq && Buffer.compare(local.value, value) !== 0) {
+          cb(Error('ERR_INVALID_SEQ'))
+          return
+        }
+        const verified = verify(sig, msg, publicKey) &&
+          (local ? seq >= local.seq : true)
```
mafintosh: seq > local.seq

stores.js
```diff
@@ -0,0 +1,273 @@
...
+          cb(err)
+          return
+        }
+        if (found === false) cb(null, null)
+      })
+    }
+    return queryStream
+  }
+  put (value, cb) {
+    assert(typeof cb === 'function', 'Callback is required')
```
mafintosh: nit: callback should be optional

mafintosh:
LGTM on tests 🎉

davidmarkclements:
tests incoming for whole module as new PR

PR #16, opened 2019-08-08
expose node id in immutable/mutable get, closes #14

### LINKED ISSUE #14: add responding node id for immutable/mutable get

davidmarkclements:


mafintosh:
LGTM on green CI. We need to fix that passthrough hack at some point as that's gonna cause trouble at some point (ie if the stream is destroyed you still emit data)

PR #26, opened 2019-11-02
propagate mutable put update errors + store prefixing

davidmarkclements:


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "dependencies": {
     "@hyperswarm/hypersign": "^2.0.0",
-    "dht-rpc": "^4.1.6",
+    "dht-rpc": "git://github.com/davidmarkclements/dht-rpc.git#expose-rtt",
```
davidmarkclements: note: temporary, needs to be altered to new release of dht-rpc before merge of this PR, waiting for merge and release of https://github.com/mafintosh/dht-rpc/pull/13

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "dependencies": {
     "@hyperswarm/hypersign": "^2.0.0",
-    "dht-rpc": "^4.1.6",
+    "dht-rpc": "git://github.com/mafintosh/dht-rpc.git#d8d6dc3",
```
davidmarkclements: reliant on https://github.com/mafintosh/dht-rpc/pull/14 - once merged this needs to be updated to next dht-rpc release

mafintosh: fixed

stores.js
```diff
@@ -192,6 +196,15 @@ class MutableStore extends Hypersign {
     const queryStream = dht.update('mutable-store', key, {
       value, signature, seq, salt
     })
+
+    queryStream.once('warning', (err, proof) => {
+      if (proof) {
```
mafintosh: check the error message here as well

mafintosh: since the error message is coming from an untrusted source it could litterally say: "Transfer BTC to this addr to proceed" with a valid proof

stores.js
```diff
@@ -192,6 +196,15 @@ class MutableStore extends Hypersign {
...
+        const { value, signature, seq, salt } = proof
+        const msg = this.signable(value, { salt, seq })
+        const verified = verify(signature, msg, key)
```
mafintosh: Based on the error message you have to here validate that error is indeed what the remote says.

Ie. if the remote says ERR_INVALID_SEQ, you have to perform the local version of the check performed here to ensure the remote is not lying, https://github.com/hyperswarm/dht/pull/26/files#diff-1071e9eefc1ddc6530576ccf7bc9adeaR241

Same for the other error

mafintosh: Otherwise this is vuln to a preimage attack (ie send some valid proof, but attack the error message itself)

stores.js
```diff
@@ -192,6 +203,18 @@ class MutableStore extends Hypersign {
...
     })
+
+    queryStream.once('warning', (err, proof) => {
+      if (err && proof) {
+        const seqErr = maybeSeqError(seq, proof.seq, proof.value, value)
+        if (seqErr) {
+          const { value, signature, seq, salt } = proof
+          const msg = this.signable(value, { salt, seq })
+          const verified = verify(signature, msg, key)
+          if (verified) queryStream.destroy(err)
```
mafintosh: seqErr

davidmarkclements:
```suggestion
          if (verified) queryStream.destroy(seqErr)
```

PR #29, opened 2020-01-19
improve tests

davidmarkclements:


mafintosh:
@davidmarkclements i realise now we never merged this. you remember if this is still up to date?

Issue #11, opened 2019-07-16
Allow setting a port for the DHT when used as binary


martinheidegger:
When starting the dht it will use a given port to connect. However, by adding a `dht.listen(port)` [here](https://github.com/hyperswarm/dht/blob/153d0fb498cc8ba106e64ba23612e28ee250a447/bin.js#L4) it should be possible to use a predictable port.

martinheidegger:
Found its way to the hyperswarm cli: https://github.com/hyperswarm/cli

PR #31, opened 2020-03-31
Add --port= to bin

da2x:
Resolves issue #11.


martinheidegger:
@da2x Thank you for working on this. I tried it out a bit (before seeing this PR) and came up with a slightly different approach in #34. The most important difference being that I create the socket before starting a query on the dht. This is probably important but I am not very familiar with firewall/network behaviour with udp.

mafintosh:
Ya --port is in the swarm cli now :)

PR #32, opened 2020-04-02
Make bootstrap servers faster and more resilient

da2x:
bootstrap3.hyperdht.org. is a a CNAME to hyperswarm.beakerbrowser.com.

Using the .com directly is faster (one fewer DNS lookup) and it makes the bootstrap servers more resilient and diverse (different TLDs, domains, DNS, and cloud infrastructure).

Assuming @pfrazee intends to maintain hyperswarm.beakerbrowser.com indefinitely.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,7 @@ const guardTimeout = require('guard-timeout')
 const DEFAULT_BOOTSTRAP = [
   'bootstrap1.hyperdht.org:49737',
   'bootstrap2.hyperdht.org:49737',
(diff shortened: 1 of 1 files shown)


mafintosh:
cc @pfrazee if you wanna change that.

I don’t think this will improve perf that much as the bootstrappers are seldomly the bottleneck in practice

pfrazee:
I'm fine with it

pfrazee:
I will be demanding a pay raise, however

PR #33, opened 2020-04-03
Add typings to hyperswarm/dht

JonasKruckenberg:
Adds typescript typings as discussed in hyperswarm/hyperswarm#55
Intellisense does not really work as extending mafintosh/dht-rpc does not have typings yet


mafintosh:
Thanks, new version out but down for typings for that!

PR #34, opened 2020-04-19
Support for port/address cli parameter

### LINKED ISSUE #11: Allow setting a port for the DHT when used as binary

martinheidegger:
To listen to a concrete address and port before the bootstrap operation we need to connect to a socket before starting the dht.

Closes #11


martinheidegger:
Closing in favour of https://github.com/hyperswarm/cli/pull/4

PR #38, opened 2020-04-30
Remove erroneous section from README.

aureooms:
There is already a section for node.immutable.get with correct information.


mafintosh:
Thanks!

PR #40, opened 2021-06-02
v5

mafintosh:
- [x] make it work
- [x] migrate immutable/mutable apis from v4 (cc @davidmarkclements)
- [ ] migrate tests from v4
- [ ] migrate examples from v4
- [x] docs


lib/holepuncher.js
```diff
@@ -0,0 +1,288 @@
...
+
+    function onconnection (rawSocket) {
+      if (self._resolveConnection === null) {
+        self._stop(null)
+        rawSocket.on('error', noop)
+        rawSocket.destroy()
+        return
+      }
+
+      const socket = this
```
davidmarkclements, line 50: is that the same as `self.socket`? if so maybe this could be taken out of the constructor and made a private method?

lib/holepuncher.js
```diff
@@ -0,0 +1,288 @@
...
+    this._stopped = false
+    this._isClient = false
+
+    // async states
+    this._openingSessions = null
+    this._holepunching = null
+    this._resolveConnection = null
+    this._onconnection = onconnection
+
+    this.connected = new Promise((resolve) => { this._resolveConnection = resolve })
```
davidmarkclements, line 40: mentioned already, but could be good to find an alternative if not too problematic elsewhere

lib/holepuncher.js
mafintosh: if would easy if that was true wouldn't it? 🤣 This is just my crazy way of reusing the event handler everywhere and find out which socket fired it using a `this` ref in there.

lib/holepuncher.js
```diff
@@ -0,0 +1,288 @@
...
+  }
+}
+
+function isFriendly (type) {
+  return type !== NAT_UNKNOWN && type !== NAT_PORT_RANDOMIZED
+}
+
+function noop () {}
+
+function holepunch (socket, addr, lowTTL) {
```
davidmarkclements, line 248: wdyt to:

```js
const send = util.promisify((socket, type, a, b, port, host, cb) => socket.send(type, a, b, port, host, cb))
async function holepunch (socket, addr, lowTTL) {
  try {
    if (lowTTL) socket.setTTL(HOLEPUNCH_TTL)
    await send(socket, HOLEPUNCH, 0, 1, addr.port, addr.host)
    return true
  } catch {
    return false
  } finally {
    if (lowTTL) socket.setTTL(DEFAULT_TTL)
  }
}
```

lib/holepuncher.js
davidmarkclements: hahah I see - I'd say leave for now and then we re-assess after full test cov

PR #41, opened 2021-06-03
WIP dht storage

davidmarkclements:
todo: 

- [x] convert hypersign signable to use compact-encoding
- [x] move hypersign.sign into this repo
- [x] tests


index.js
```diff
@@ -569,6 +632,30 @@ function allowAll () {
...
 
+function mapImmutable (node) {
+  const query = this
+  return {
+    id: node.id,
+    value: node.value,
+    token: node.token,
+    from: node.from,
+    to: node.to,
+    peers: query.dht.peers
```
mafintosh, line 643: not needed here

index.js
```diff
@@ -569,6 +632,30 @@ function allowAll () {
...
+
+function mapMutable (node) {
+  const query = this
+  return {
+    id: node.id,
+    value: node.value,
+    token: node.token,
+    from: node.from,
+    to: node.to,
+    peers: query.dht.peers
```
mafintosh, line 655: also no peers here as well

davidmarkclements:
*NOTE* - minor sec fixes required in hypersign

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+    const key = Buffer.allocUnsafe(32)
+    sodium.crypto_generichash(key, value)
+    const query = this.query(key, 'immutable_get', null, {
+      map: mapImmutable,
+      commit (node, dht) {
+        return dht.request(key, 'immutable_put', value, node.from, { map: mapImmutable, token: node.token })
+      }
+    })
+    await query.finished()
+    return { key }
```
mafintosh, line 222: include closestNodes from query here also (useful to refresh the put)

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+    const check = Buffer.allocUnsafe(32)
+    for await (const node of query) {
+      const { value } = node
+      sodium.crypto_generichash(check, value)
+      if (check.equals(key)) return node
+    }
+    throw Error('not found')
+  }
+
+  async immutablePut (value) {
```
mafintosh, line 208: add opts = {}

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+    if (Buffer.isBuffer(value) === false) throw new Error('value must be a buffer')
+    if (value.length > PUT_VALUE_MAX_SIZE) {
+      throw new Error(`Value size must be <= ${PUT_VALUE_MAX_SIZE}`)
+    }
+    const key = Buffer.allocUnsafe(32)
+    sodium.crypto_generichash(key, value)
+    const query = this.query(key, 'immutable_get', null, {
```
mafintosh, line 215: spread in opts from above

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+    if (salt) {
+      if (Buffer.isBuffer(salt) === false) throw new Error('salt must be a buffer')
+      if (salt.length > 64) throw new Error('salt size must be no greater than 64 bytes')
+    }
+
+    const request = cenc.encode(messages.mutableReq, { salt, seq })
+
+    const query = this.query(key, 'mutable_get', request, { map: mapMutable })
+    const userSeq = seq
+    let topSeq = seq
```
mafintosh, line 237: would it make sense to rename seq in the opts to `gt` and `gte` ? (that's how our db abstractions name the limits)

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+    let result = null
+    for await (const node of query) {
+      const { value: item, id, ...meta } = node
+      const { value, signature, seq: storedSeq, publicKey } = cenc.decode(messages.mutableRes, item)
```
mafintosh, line 241: you need to try catch continue this in case a peer sends back bogus data

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+    const { publicKey } = keyPair
+    const key = Buffer.allocUnsafe(32)
+    sodium.crypto_generichash(key, publicKey)
+    const msg = cenc.encode(messages.mutableRes, {
+      value, signature, seq, salt, publicKey
+    })
+    const query = this.query(key, 'mutable_get', null, {
+      map: mapMutable,
+      commit (node, dht) {
+        return dht.request(key, 'mutable_put', msg, node.from, { map: mapMutable, token: node.token })
```
mafintosh, line 276: map doesn't do anything here

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+      }
+    })
+    await query.finished()
+    return { key, signature, seq, salt }
```
mafintosh, line 280: return closestNodes here also

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+      if (Buffer.isBuffer(publicKey) === false) throw new Error('keyPair.publicKey is required')
+      if (secretKey) throw new Error('only opts.signature OR opts.keyPair.secretKey should be supplied')
+    }
+    const { publicKey } = keyPair
+    const key = Buffer.allocUnsafe(32)
+    sodium.crypto_generichash(key, publicKey)
+    const msg = cenc.encode(messages.mutableRes, {
+      value, signature, seq, salt, publicKey
+    })
+    const query = this.query(key, 'mutable_get', null, {
```
mafintosh, line 273: forward opts.closestNodes here

lib/messages.js
```diff
@@ -103,6 +103,49 @@ exports.connectRelay = {
   }
 }
 
+exports.mutableReq = {
```
mafintosh, line 106: nit: Request instead of Req as it's an exposes property

lib/messages.js
```diff
@@ -103,6 +103,49 @@ exports.connectRelay = {
...
+  },
+  decode (state) {
+    return {
+      seq: cenc.uint.decode(state),
+      salt: cenc.buffer.decode(state)
+    }
+  }
+}
+
+exports.mutableRes = {
```
mafintosh, line 123: nit: Response instead of Res as it's an exposed property

lib/messages.js
```diff
@@ -103,6 +103,49 @@ exports.connectRelay = {
...
+    cenc.fixed32.encode(state, m.publicKey)
+    cenc.buffer.encode(state, m.salt)
+  },
+  decode (state) {
+    return {
+      seq: cenc.uint.decode(state),
+      value: cenc.buffer.decode(state),
+      signature: cenc.buffer.decode(state),
+      publicKey: cenc.fixed32.decode(state),
+      salt: cenc.buffer.decode(state)
```
mafintosh, line 144: salt is provided by the user in the response so it should not be sent over in the response

lib/messages.js
```diff
@@ -103,6 +103,49 @@ exports.connectRelay = {
...
+    cenc.buffer.encode(state, m.value)
+    cenc.buffer.encode(state, m.signature)
+    cenc.fixed32.encode(state, m.publicKey)
+    cenc.buffer.encode(state, m.salt)
+  },
+  decode (state) {
+    return {
+      seq: cenc.uint.decode(state),
+      value: cenc.buffer.decode(state),
+      signature: cenc.buffer.decode(state),
```
mafintosh, line 142: signature is fixed64

mafintosh: same in above enc/preenc

lib/persistent.js
```diff
@@ -31,6 +35,53 @@ module.exports = class PersistentNode {
     return this.dht.id.equals(TMP)
   }
 
+  onget (req, { mutable }) {
+    if (!req.target) return req.reply(null)
+    const store = mutable ? this.mutables : this.immutables
+    if (mutable) {
+      const k = req.target.toString('hex')
+      const value = store.get(k)
+      if (!value) return req.reply(Buffer.from('')) // empty response so initial puts can still find nodes
```
mafintosh, line 44: empty value not required, just reply(null)

lib/persistent.js
```diff
@@ -31,6 +35,53 @@ module.exports = class PersistentNode {
...
+      return req.reply(value)
+    }
+
+    const { salt, seq } = cenc.decode(messages.mutableReq, req.value)
+    const k = salt ? req.target.toString('hex') + salt.toString('hex') : req.target.toString('hex')
+    const value = store.get(k)
+    if (!value) return req.reply(Buffer.from('')) // empty response so initial puts can still find nodes
```
mafintosh, line 51: same as above, just do reply(null)

lib/persistent.js
```diff
@@ -31,6 +35,53 @@ module.exports = class PersistentNode {
...
+    const local = cenc.decode(messages.mutableRes, value)
+    if (local.seq < seq) return req.reply(null)
+    return req.reply(value)
+  }
+
+  onput (req, { mutable }) {
+    if (!req.target || !req.value) return req.reply(null)
```
mafintosh, line 58: also add !req.token

lib/persistent.js
```diff
@@ -31,6 +35,53 @@ module.exports = class PersistentNode {
...
+    const store = mutable ? this.mutables : this.immutables
+    if (mutable === false) {
+      const k = req.target.toString('hex')
+      store.set(k, req.value)
```
mafintosh, line 62: you need to verify the hash first (ie target == hash(val))

lib/persistent.js
mafintosh: In general, whenever the requester does something wrong just return instead of reply(null) - they are not following the protocol, they don't reserve a reply

lib/persistent.js
```diff
@@ -31,6 +35,53 @@ module.exports = class PersistentNode {
...
+    const key = salt
+      ? req.target.toString('hex') + salt.toString('hex')
+      : req.target.toString('hex')
+    const local = store.get(key)
+    const msg = hypersign.signable(value, { salt, seq })
+    const verified = sodium.crypto_sign_verify_detached(signature, msg, publicKey)
+    if (verified === false) return // timeout
+    if (local) {
+      if (local.value && local.seq === seq && Buffer.compare(value, local.value) !== 0) {
+        return req.error(INVALID_SEQ)
```
mafintosh, line 76: nit: this is SEQ_REUSE instead of invalid seq isn't it?

index.js
```diff
@@ -569,6 +657,28 @@ function allowAll () {
   return true
 }
 
+function mapImmutable (node) {
```
andrewosh, line 660: Would probably move the decoding into these `map` methods

index.js
```diff
@@ -192,6 +193,93 @@ module.exports = class HyperDHT extends DHT {
...
+    }
+    return result
+  }
+
+  async mutablePut (value, opts = {}) {
+    if (Buffer.isBuffer(value) === false) throw new Error('value must be a buffer')
+    if (value.length > PUT_VALUE_MAX_SIZE) {
+      throw new Error(`Value size must be <= ${PUT_VALUE_MAX_SIZE}`)
+    }
+    const { seq = 0, salt = null, keyPair, signature = hypersign.sign(value, { ...opts, keypair: keyPair }) } = opts
```
mafintosh, line 259: this one is bit out of my paygrade i think, what's going on?

index.js
davidmarkclements: can I just spread the opts as in the other case

lib/messages.js
davidmarkclements: I called it mutablePut at first but it looks weird because it's used in `onget` and `mutableGet` - that's why I went to Req

lib/persistent.js
davidmarkclements: I did do this and it meant that I couldn't get any nodes for an initial put, I'll try it again, perhaps it was another reason

index.js
davidmarkclements: ahha, it's just auto generating a signature from the keypair if signature is not supplied 

I'll move the signing into dht and then we can revisit

index.js
davidmarkclements: moved try/catch in mapMutable function

index.js
mafintosh: ya lets keep at seq

lib/persistent.js
```diff
@@ -31,6 +35,57 @@ module.exports = class PersistentNode {
...
+      if (hash.equals(req.target) === false) return // timeout
+      const k = req.target.toString('hex')
+      store.set(k, req.value)
+      return req.reply(null)
+    }
+    const { seq, value, signature, publicKey } = cenc.decode(messages.mutable, req.value)
+    const hash = Buffer.alloc(32)
+    sodium.crypto_generichash(hash, publicKey)
+    if (hash.equals(req.target) === false) return // timeout
+    if (!value || !signature) return // timeout
```
mafintosh, line 71: signature cannot be falsy (fixed64)

lib/persistent.js
```diff
@@ -31,6 +35,57 @@ module.exports = class PersistentNode {
...
+    if (mutable === false) {
+      const hash = Buffer.alloc(32)
+      sodium.crypto_generichash(hash, req.value)
+      if (hash.equals(req.target) === false) return // timeout
+      const k = req.target.toString('hex')
+      store.set(k, req.value)
+      return req.reply(null)
+    }
+    const { seq, value, signature, publicKey } = cenc.decode(messages.mutable, req.value)
+    const hash = Buffer.alloc(32)
```
mafintosh, line 68: you can allocUnsafe here

mafintosh (approved):

PR #45, opened 2021-07-06
Line re-ordering to fix temporal dead zone error

4c656554:
Definition of `value` and `error` in `connectRaw(publicKey, opts = {})` must be before reference to `ontimeout`.

Else a  "temporal dead zone error" might occur:
https://stackoverflow.com/a/62203661

See issue 44:

 hyperswarm#44


mafintosh:
Fixed in master, thanks!

mafintosh:
And released

Issue #44, opened 2021-06-28
There appears to be a line order error in index.js


lanmower:
There was an uncaught error ReferenceError: Cannot access 'error' before initialization
    at Timeout.ontimeout [as _onTimeout] (C:\app\dev\test\node_modules\@hyperswarm\dht\index.js:216:7)
    at listOnTimeout (internal/timers.js:555:17)
    at processTimers (internal/timers.js:498:7)

4c656554:
I get this too, with slightly different line numbers:

```
at listOnTimeout (internal/timers.js:554:17)
at processTimers (internal/timers.js:497:7)
```

PR #47, opened 2021-07-07
Declare query and holepunch before reference to ontimeout

4c656554:
Re issue #44

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,7 +131,7 @@ module.exports = class HyperDHT extends DHT {
     await this.sampledNAT()
 
     const addr = this.remoteAddress()
-    const holepunch = new Holepuncher(addr)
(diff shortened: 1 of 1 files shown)


mafintosh:
In ontimeout you need to check if holepunch and query are set as well before destroying them, ie

```
if (holepunch) holepunch.destroy()
```

4c656554:
Made a new commit...

mafintosh:
Awesome, will merge on green ci :)

index.js
```diff
@@ -116,6 +116,8 @@ module.exports = class HyperDHT extends DHT {
 
   async connectRaw (publicKey, opts = {}) {
     let error = null
+    let query
+    let holepunch
```
mafintosh, line 120: standard wants you to initialise these to null

PR #64, opened 2021-11-24
move to the lts ns scheme for sigs/caps

mafintosh:
(in a PR as this is currently not backwards compat, so just need to make sure it's rolled up in one go on the testnet)


mafintosh:
Will merge this when we move off the testnet

mafintosh:
lts scheme in master now

PR #65, opened 2021-11-29
Add browser stub to improve bundling and switch to `b4a`

kasperisager:


kasperisager:
I should probably add the `keyPair()` function which is commonly used and works just fine in a browser.

index.js
```diff
@@ -183,7 +185,7 @@ class HyperDHT extends DHT {
   async mutableGet (publicKey, opts = {}) {
     opts = { ...opts, map: mapMutable }
 
-    const target = Buffer.alloc(32)
+    const target = b4a.alloc(32)
```
mafintosh, line 188: unsafe is fine here actually, missed that before

PR #67, opened 2021-12-20
Add `sign` option to APIs that require signing

kasperisager:
This pull request introduces a new `sign` option to APIs that require signing, such as `node.announce()` and `node.mutablePut()`. If given, this property is expected to be an optionally async function that signs the given buffer.

It still lacks tests and docs as I wanted to get some input on the API first.

### DIFF
--- a/lib/persistent.js
+++ b/lib/persistent.js
@@ -220,28 +220,24 @@ module.exports = class Persistent {
     req.reply(null)
   }
 
-  static signMutable (seq, value, secretKey) {
-    const signature = Buffer.allocUnsafe(64)
+  static signMutable (seq, value, secretKeyOrSigner) {
     const signable = Buffer.allocUnsafe(32)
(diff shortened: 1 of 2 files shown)


lib/persistent.js
```diff
@@ -220,28 +220,24 @@ module.exports = class Persistent {
...
+  static signAnnounce (target, token, id, ann, secretKeyOrSigner) {
+    const signable = annSignable(target, token, id, ann, NS.ANNOUNCE)
+    return sign(signable, secretKeyOrSigner)
   }
 
-  static signUnannounce (target, token, id, ann, secretKey) {
-    const signature = Buffer.allocUnsafe(64)
-    sodium.crypto_sign_detached(signature, annSignable(target, token, id, ann, NS.UNANNOUNCE), secretKey)
-    return signature
+  static signUnannounce (target, token, id, ann, secretKeyOrSigner) {
```
mafintosh, line 238: instead of having the argument here be polymorphic can we always instead expect it to be a function and then just have a default one?

mafintosh: Same for the other ones

PR #68, opened 2022-01-05
Support externally managed secret keys

kasperisager:
This pull request introduces support for externally managed secret keys to allow a relaying DHT node to act on behalf of another relayed node. A summary of the changes:

- `DHT.createServer()` and `DHT.connect()` now accept two new options, `createHandshake` and `createSecretStream`, to allow callers to perform Noise handshaking and stream setup themselves. To ensure that the relaying DHT node can still hole punch on behalf of the relayed node, `handshake.final()` is now required to return a `holepunchSecret` key, but may leave out the Noise keys.

- `server.listen()` now accepts two new options, `signAnnounce` and `signUnannounce`, to allow callers to perform signing themselves.

- `DHT.lookupAndUnannounce()`, `DHT.unannounce()`, and `DHT.announce()` now accept a new option, `signUnannounce`, to allow callers to perform signing themselves. `DHT.announce()` also accepts a `signAnnounce` option and `DHT.mutablePut()` accepts a `signMutable` option.

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,12 +86,15 @@ class HyperDHT extends DHT {
       this._persistent.unannounce(target, keyPair.publicKey)
     }
 
-    const commit = async (reply, dht, query) => {
+    opts = { ...opts, map, commit }
+    return this.query({ target, command: COMMANDS.LOOKUP, value: null }, opts)
+
+    async function commit (reply, dht, query) {
       await Promise.all(unannounces) // can never fail, caught below
       return userCommit(reply, dht, query)
     }
 
-    const map = (reply) => {
+    function map (reply) {
       const data = mapLookup(reply)
 
       if (!data || !data.token) return data
@@ -116,10 +119,6 @@ class HyperDHT extends DHT {
 
       return data
     }
-
-    opts = { ...opts, map, commit }
-
-    return this.query({ target, command: COMMANDS.LOOKUP, value: null }, opts)
   }
 
   unannounce (target, keyPair, opts = {}) {
@@ -129,8 +128,14 @@ class HyperDHT extends DHT {
   announce (target, keyPair, relayAddresses = [], opts = {}) {
     const signAnnounce = opts.signAnnounce || Persistent.signAnnounce
 
-    const commit = (reply, dht) => {
-      return this._requestAnnounce(
+    opts = { ...opts, commit }
+
+    return opts.clear
+      ? this.lookupAndUnannounce(target, keyPair, opts)
+      : this.lookup(target, opts)
+
+    function commit (reply, dht) {
+      return dht._requestAnnounce(
         keyPair,
         dht,
         target,
@@ -140,12 +145,6 @@ class HyperDHT extends DHT {
         signAnnounce
       )
     }
-
-    opts = { ...opts, commit }
-
-    return opts.clear
-      ? this.lookupAndUnannounce(target, keyPair, opts)
-      : this.lookup(target, opts)
   }
 
   async immutableGet (target, opts = {}) {
--- a/index.js
+++ b/index.js
@@ -80,6 +80,7 @@ class HyperDHT extends DHT {
     const unannounces = []
     const dht = this
     const userCommit = opts.commit || noop
+    const signUnannounce = opts.signUnannounce || Persistent.signUnannounce
 
     if (this._persistent !== null) { // unlink self
       this._persistent.unannounce(target, keyPair.publicKey)
@@ -114,7 +115,7 @@ class HyperDHT extends DHT {
         signature: null
       }
 
-      unann.signature = Persistent.signUnannounce(target, token, from.id, unann, keyPair.secretKey)
+      unann.signature = signUnannounce(target, token, from.id, unann, keyPair.secretKey)
 
       const value = c.encode(m.announce, unann)
       unannounces.push(dht.request({ token, target, command: COMMANDS.UNANNOUNCE, value }, from).catch(noop))
@@ -128,6 +129,8 @@ class HyperDHT extends DHT {
   }
 
   announce (target, keyPair, relayAddresses = [], opts = {}) {
+    const signAnnounce = opts.signAnnounce || Persistent.signAnnounce
+
     opts = { ...opts, commit }
 
     return opts.clear
@@ -145,7 +148,7 @@ class HyperDHT extends DHT {
         signature: null
       }
 
-      ann.signature = Persistent.signAnnounce(target, token, from.id, ann, keyPair.secretKey)
+      ann.signature = signAnnounce(target, token, from.id, ann, keyPair.secretKey)
 
       const value = c.encode(m.announce, ann)
       return dht.request({ token, target, command: COMMANDS.ANNOUNCE, value }, from)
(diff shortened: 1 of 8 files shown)


lib/noise-wrap.js
```diff
@@ -37,11 +38,16 @@ module.exports = class NoiseWrap {
...
+    const holepunchCapability = Buffer.allocUnsafe(32)
+
+    sodium.crypto_generichash(holepunchCapability, this.handshake.hash, NS.PEER_HOLEPUNCH)
+
     return {
       isInitiator: this.isInitiator,
       publicKey: this.keyPair.publicKey,
       remotePublicKey: this.remotePublicKey,
       remoteId: NoiseSecretStream.id(this.handshake.hash, !this.isInitiator),
+      holepunchCapability,
```
mafintosh, line 50: holepunchSecret

kasperisager:
@davidmarkclements The tests are failing on Windows as of Brittle 2.0.2. Do you know what might be causing this?

kasperisager:
@mafintosh This is ready for another look.

lib/server.js
```diff
@@ -105,38 +114,48 @@ module.exports = class Server extends EventEmitter {
...
-    const fw = this.firewall(handshake.remotePublicKey, remotePayload, clientAddress)
-
     const hs = {
       round: 0,
       reply: null,
       pair: null,
-      protocols: remotePayload.protocols & this._protocols,
-      firewalled: (!fw || !fw.then) ? Promise.resolve(fw) : fw.catch(toTrue),
+      protocols: this._protocols,
+      firewalled: false,
```
mafintosh, line 126: init this to true, just to be on the safe side

lib/noise-wrap.js
```diff
@@ -16,32 +17,30 @@ module.exports = class NoiseWrap {
...
-    let payload = null
-
-    try {
-      payload = c.decode(messages.noisePayload, this.handshake.recv(buf))
-    } catch {
-      return null
-    }
-
+  async recv (buf) {
+    const payload = c.decode(messages.noisePayload, this.handshake.recv(buf))
```
mafintosh, line 26: nice, good change!

lib/announcer.js
```diff
@@ -6,18 +6,20 @@ const Persistent = require('./persistent')
...
 module.exports = class Announcer {
-  constructor (dht, keyPair, target) {
+  constructor (dht, keyPair, target, record, opts = {}) {
     this.dht = dht
     this.keyPair = keyPair
     this.target = target
     this.relays = []
     this.stopped = false
-    this.record = c.encode(m.peer, { publicKey: this.keyPair.publicKey, relayAddresses: [] })
+    this.record = record
```
mafintosh, line 15: is this change still needed now? otherwise let's move it to another pr or just keep it as it was

index.js
```diff
@@ -208,11 +211,13 @@ class HyperDHT extends DHT {
...
 
   async mutablePut (keyPair, value, opts = {}) {
+    const signMutable = opts.signMutable || Persistent.signMutable
+
     const target = Buffer.allocUnsafe(32)
     sodium.crypto_generichash(target, keyPair.publicKey)
 
     const seq = opts.seq || 0
-    const signature = Persistent.signMutable(seq, value, keyPair.secretKey)
+    const signature = signMutable(seq, value, keyPair.secretKey)
```
mafintosh, line 220: this needs async support doesnt it?

index.js
```diff
@@ -145,7 +148,7 @@ class HyperDHT extends DHT {
         signature: null
       }
 
-      ann.signature = Persistent.signAnnounce(target, token, from.id, ann, keyPair.secretKey)
+      ann.signature = signAnnounce(target, token, from.id, ann, keyPair.secretKey)
```
mafintosh, line 151: async now? note the commit method can be async to help with that (it expects a promise returned)

index.js
```diff
@@ -114,7 +115,7 @@ class HyperDHT extends DHT {
         signature: null
       }
 
-      unann.signature = Persistent.signUnannounce(target, token, from.id, unann, keyPair.secretKey)
+      unann.signature = signUnannounce(target, token, from.id, unann, keyPair.secretKey)
```
mafintosh, line 118: async also yea? note that this is doable if you just move this logic to a _requestUnnnnounce method or something that is async and call that instead of dht.request below (the unannounces is a list of promises)

lib/connect.js
```diff
@@ -17,7 +17,7 @@ module.exports = function connect (dht, publicKey, opts = {}) {
     dht,
     round: 0,
     target: hash(publicKey),
-    handshake: new NoiseWrap(keyPair, publicKey),
+    handshake: (opts.handshake || defaultHandshake)(keyPair, publicKey),
```
mafintosh, line 20: nit: createHandshake?

lib/connect.js
```diff
@@ -182,7 +182,7 @@ async function connectThroughNode (c, node) {
   const { serverAddress, clientAddress, relayed, noise } = await c.dht._router.peerHandshake(c.target, { noise: c.request }, node.from)
   if (isDone(c) || c.connect) return
 
-  const payload = c.handshake.recv(noise)
+  const payload = await c.handshake.recv(noise)
   if (!payload) return
```
mafintosh, line 186: check isDone(c) here also like above, since this follows an await now

lib/connect.js
```diff
@@ -170,7 +170,7 @@ async function connectThroughNode (c, node) {
     const addr = c.dht._sockets.remoteServerAddress()
 
     c.firewall = c.dht.firewalled ? FIREWALL.UNKNOWN : FIREWALL.OPEN
-    c.request = c.handshake.send({
+    c.request = await c.handshake.send({
```
mafintosh, line 173: this needs an extra flag on c now, requesting: true|false, because we guard above using the c.request property if we should issue the handshake at all, and the await breaks that

lib/noise-wrap.js
```diff
@@ -22,26 +23,24 @@ module.exports = class NoiseWrap {
...
     return payload
   }
 
   final () {
     if (!this.handshake.complete) throw new Error('Handshake did not finish')
 
+    const holepunchSecret = Buffer.allocUnsafe(32)
+
+    sodium.crypto_generichash(holepunchSecret, this.handshake.hash, NS.PEER_HOLEPUNCH)
+
```
mafintosh, line 37: just a mental note that we might want to have a static function or someway of generating this outside the module now we need it returned from final :)

lib/server.js
```diff
@@ -91,6 +94,8 @@ module.exports = class Server extends EventEmitter {
       onpeerholepunch: this._onpeerholepunch.bind(this)
     })
 
+    this._listening = true
```
mafintosh, line 97: you need to set this to false also behlow in the catch

lib/server.js
```diff
@@ -105,38 +110,48 @@ module.exports = class Server extends EventEmitter {
...
     }
-    if (this.closed) {
+
+    hs.protocols &= remotePayload.protocols
+
+    try {
+      hs.firewalled = await this.firewall(handshake.remotePublicKey, remotePayload, clientAddress)
+    } catch (err) {
+      safetyCatch(err)
+      hs.firewalled = true
```
mafintosh, line 145: not needed now that it's initialised to true

kasperisager:
I broke the tests 💥

lib/connect.js
```diff
@@ -165,12 +166,13 @@ async function findAndConnect (c, opts) {
...
 async function connectThroughNode (c, node) {
-  if (!c.request) {
+  if (!c.requesting) {
     // If we have a stable server address, send it over now
     const addr = c.dht._sockets.remoteServerAddress()
 
     c.firewall = c.dht.firewalled ? FIREWALL.UNKNOWN : FIREWALL.OPEN
-    c.request = c.handshake.send({
+    c.requesting = true
+    c.request = await c.handshake.send({
```
mafintosh, line 175: Last thing, we should check isDone after the send also

PR #73, opened 2022-01-17
Add browser.js to avoid nodejs dependencies

Nazeh:
Added dependency: `b4a`
Could have refactored `createKeyPair` to be shared between`index.js` and `browser.js` but I wanted to keep changes to `index.js` minimal.


kasperisager:
Duplicate of #65

PR #78, opened 2022-02-09
fix: fix address undefined error

gmaclennan:
I came across this error when creating a DHT on a local network. If `addr.host` is a private or bogon address, then `address` can end up being null (because for some reason `reply` is `null` the first time that `connect()` is called), which results in an uncaught error.

### DIFF
--- a/lib/socket-pairer.js
+++ b/lib/socket-pairer.js
@@ -427,6 +427,10 @@ class Pair {
 
     if (!address) address = from
 
+    // If `from` is null, and `addr.host` is private or bogon, then `address`
+    // will be undefined here, so the next line will throw an error.
+    if (!address) return
+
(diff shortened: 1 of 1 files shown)


lib/socket-pairer.js
```diff
@@ -427,6 +427,10 @@ class Pair {
 
     if (!address) address = from
 
+    // If `from` is null, and `addr.host` is private or bogon, then `address`
+    // will be undefined here, so the next line will throw an error.
+    if (!address) return
```
mafintosh, line 432: return false

mafintosh:
I think the from === null thing is a bug actually

gmaclennan:
Yeah thought it might be. I'm afraid I don't really know enough about the internals to fix that one.

mafintosh:
Should be fixed in master now

PR #82, opened 2022-03-29
Switch to `udx-native`

kasperisager:
This PR replaces the networking layer with https://github.com/mafintosh/libudx.


lib/socket-pairer.js
```diff
@@ -487,31 +478,46 @@ class Pair {
...
-      self._ontcpconnection(rawSocket, null, false)
+    function connect () {
+      rawSocket.pause()
+      self._onudxconnection(rawSocket, null, false, null)
+    }
+
+    function close () {
+      // TODO: when not deferred, the socket is closed before the other side
+      // has its `end` message ack'ed; how can we handle this?
+      setTimeout(() => socket.close())
```
kasperisager, line 512: Tracking this here: https://github.com/mafintosh/libudx/pull/36

test/connections.js
```diff
@@ -316,9 +316,12 @@ test('server responds and immediately ends, multiple connects', async function (
   for (let i = n; i > 0; i--) {
     const socket = b.connect(server.publicKey)
 
-    socket.on('open', () => {
-      if (--n === 0) lc.pass()
-    })
```
kasperisager, line 322: The socket was throwing an `ECONNRESET` error after moving to UDX. This is perfectly aligned with TCP behaviour though, so that seems intended.

test/connections.js
```diff
@@ -266,25 +264,26 @@ test('client choosing to abort holepunch', async function (t) {
...
+  const socket = dgram.createSocket('udp4')
+  socket.send('hi', node.address().port)
+  socket.close()
 
-  await new Promise((resolve) => sock.on('close', resolve))
+  await new Promise((resolve) => socket.on('close', resolve))
   t.pass('did not crash')
 })
 
 test('half open', async function (t) {
```
kasperisager, line 278: This case is currently failing intermittently due to congestion.

index.js
```diff
@@ -340,6 +347,14 @@ class HyperDHT extends DHT {
       value
     }, from)
   }
+
+  _availableStreamId () {
+    while (true) {
+      const id = (Math.random() * 0xffffffff) >>> 0
```
mafintosh, line 353: hate to be that guy, but this should be `random() * 0x100000000`

mafintosh: I feel like this should be a public method on the pool also instead now

kasperisager: Always off by 1 🙈

lib/socket-pool.js
```diff
@@ -0,0 +1,181 @@
+const udx = require('udx-native')
+
+module.exports = class SocketPool {
+  constructor (dht) {
+    this._dht = dht
+
+    this._sockets = new Set()
+    this._free = new Set()
```
mafintosh, line 8: this should be an array imo so we can track causality later on

lib/socket-pool.js
```diff
@@ -0,0 +1,181 @@
...
+      port
+    } = options
+
+    let socket = null
+
+    if (port) socket = this._getByPort(port)
+
+    // If we don't have a socket yet, try to get a free one
+    if (!socket) {
+      [socket] = this._free
```
mafintosh, line 51: if/when free was an array: `socket = this._free.pop()`

lib/socket-pool.js
```diff
@@ -0,0 +1,181 @@
...
+  constructor (dht) {
+    this._dht = dht
+
+    this._sockets = new Set()
+    this._free = new Set()
+    this._pairers = new Map()
+    this._ids = new Set()
+
+    // Indexes
+    this._byPort = new Map()
```
mafintosh, line 13: don't think we need a port index actually, but we can keep it until we know for sure

lib/socket-pool.js
kasperisager: What exactly would we track with an array?

lib/socket-pool.js
kasperisager: We shouldn't need one, no.

lib/stream-set.js
```diff
@@ -0,0 +1,29 @@
+module.exports = class StreamSet {
+  constructor (dht) {
+    this._dht = dht
+
+    this._streams = new Map()
+  }
+
+  create (opts) {
```
mafintosh, line 8: mb .addStream or .add ?

lib/stream-set.js
```diff
@@ -0,0 +1,29 @@
+module.exports = class StreamSet {
```
mafintosh, line 1: could also be RawStreamSet to make it clear it isn't the encrypted streams

mafintosh:
@kasperisager pushed my changes to origin/udx. can we update this PR to point to that?

kasperisager:
GitHub only allows changing the target branch, not the source branch. I can open a new PR though.

mafintosh:
@kasperisager really! can you make a new PR then?

kasperisager:
Superseded by #83.

PR #83, opened 2022-05-04
Switch to `udx-native`

kasperisager:
Supersedes #82. Depends on mafintosh/dht-rpc#40.


lib/connect.js
```diff
@@ -27,29 +24,45 @@ module.exports = function connect (dht, publicKey, opts = {}) {
     handshake: (opts.createHandshake || defaultCreateHandshake)(keyPair, publicKey),
     request: null,
     requesting: false,
-    protocols: PROTOCOL.TCP | PROTOCOL.UTP,
+    protocols: PROTOCOL.UDX,
```
kasperisager, line 27: @mafintosh Do we still need this?

lib/socket-pool.js
```diff
@@ -29,6 +29,10 @@ module.exports = class SocketPool {
     this._free.delete(holder)
   }
 
+  _release (holder) {
+    if (this._sockets.has(holder)) this._free.add(holder)
```
mafintosh, line 33: shouldn't we close the socket here if idle also? otherwise who closes these

lib/connect.js
mafintosh: no, but let's save this for a "handshake bikeshed session 2022 edition"

mafintosh:
Landed in master, with the additional handshake tweak from our other PR \o/ - leaving this branch around as we have some tests using it

PR #87, opened 2022-06-27
noise wrap passes optional curve to noise handshake async recv

rafapaezbas:
The noise handshake used in peers connection could use a custom Diffie-Hellman implementation. The Noisewarp constructor would benefit from passing the curve to the NoiseHandshake constructor making it more flexible. Besides, the implementation could be asynchronous so the receive method must await in order to decode the received buffer.

See this example:

``` js
const createHandshake = (keyPair, remotePublicKey) => {
	const curve = {
          dh: async (pk, seck) => customDH(pk, seck),
	  PKLEN: 32,
          ALG: 'Ed25519',
          name: 'Ed25519',
	  generateKeyPair
	}
	return new NoiseWrap(keyPair, remotePublicKey, { curve })
}

const node = new DHT()
const server = node.createServer({ createHandshake })
```

### DIFF
--- a/lib/noise-wrap.js
+++ b/lib/noise-wrap.js
@@ -10,11 +10,11 @@ const { NS } = require('./constants')
 const NOISE_PROLOUGE = NS.PEER_HANDSHAKE
 
 module.exports = class NoiseWrap {
-  constructor (keyPair, remotePublicKey) {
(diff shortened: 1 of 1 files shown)


mafintosh:
dont think this is relevant anymore, if it is please reopen

PR #89, opened 2022-07-04
Reusable sockets

mafintosh:
Allows peer to opt-in to reusing udx socket and if both peers do, they cache the route for a bit if no connection errors happens so if a subsequent connection is made it'll reuse the same route.

If the route reuse fails, it'll fallback to a full dht lookup still


lib/socket-pool.js
```diff
@@ -1,58 +1,127 @@
...
+
+    rawStream.on('error', () => {
+      this._pool.setReusable(socket, false)
+      if (!route) route = this._routes.get(id)
+      if (route && route.socket === socket) this._routes.delete(id)
+    })
+  }
+}
+
+// TODO: we should just make some "user data" object on udx to allow to attach this info
```
kasperisager, line 99: `stream.userData` and `socket.userData` are available in `udx-native` 1.2.0; should we use those?

kasperisager (changes requested):

lib/socket-pool.js
mafintosh: ya let's do that in a follow up pr

kasperisager (approved):
Looks good! Not sure if there's an easy way to test socket reuse without interrogating too many internals?

PR #90, opened 2022-07-04
Better non firewall options

mafintosh:
If a server is not firewalled (ie has a remote addr), no need to allocate holepunching sockets


kasperisager (approved):
Nice 👌

PR #93, opened 2022-07-27
Lock down IP object shapes

kasperisager:
As the DHT keeps IPv4 and IPv6 addresses separate, the `family` property introduced in https://github.com/compact-encoding/compact-encoding-net/pull/1 isn't needed for distinguishing them.

### DIFF
--- a/lib/messages.js
+++ b/lib/messages.js
@@ -1,9 +1,30 @@
 const c = require('compact-encoding')
 const net = require('compact-encoding-net')
 
-const ipv4 = net.ipv4Address
+const ipv4 = {
+  ...net.ipv4Address,
+  decode (state) {
+    const ip = net.ipv4Address.decode(state)
+    return {
+      host: ip.host,
+      port: ip.port
+    }
+  }
+}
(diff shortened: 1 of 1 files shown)


mafintosh:
nice and simple 👍

mafintosh:
@kasperisager can you check if we pass the address from on('message') directly to other apis as well, re shapes (that one has family also, which makes sense)

kasperisager:
@mafintosh It's passed on to internal APIs, but it doesn't look like the `family` property ever reaches public API surface.

kasperisager:
For this callsite, for example, I think we specifically _want_ the `family` property to ensure monomorphism: https://github.com/hyperswarm/dht/blob/737c46286670beddd53f226795f2b1e8e8c93356/lib/socket-pool.js#L15

That handler is also invoked from the DHT client and server sockets where the addresses contain the `family` property.

mafintosh:
@kasperisager ya makes sense. was referring internally where we use that address, but let's check after merge

PR #94, opened 2022-07-28
Adapt stream MTU on `connect`

kasperisager:


lib/raw-stream-set.js
```diff
@@ -18,12 +18,29 @@ module.exports = class RawStreamSet {
...
+
+    async function onconnect () {
+      const {
+        remoteHost: host,
+        remotePort: port
+      } = stream
+
+      for (let mtu = 1400; mtu > 1200; mtu -= 100) {
+        try {
+          await self._dht.ping({ host, port }, { size: mtu - 12 /* overhead */ })
```
mafintosh, line 39: pass the socket here also as options, stream.rawStream.socket so it has the correct session

kasperisager: Terribly good point 😂

lib/raw-stream-set.js
```diff
@@ -18,12 +20,35 @@ module.exports = class RawStreamSet {
...
+        try {
+          await self._dht.ping({ host, port }, {
+            size: mtu - 12, /* overhead */
+            socket: stream.socket
+          })
+
+          stream.setMTU(mtu)
+          break
+        } catch (err) {
+          safetyCatch(err)
```
mafintosh, line 49: @kasperisager can you test what happens if this runs when the socket has been closed?

mafintosh: ie, if the socket is opened and closed and then we run this ping in the background, just to make sure we don't crash becaue of that

kasperisager: I'm not sure if I can construct a case where the socket is closed before the ping is attempted 🤔 The `catch` clause is hit during the other test cases though.

kasperisager: @mafintosh Let me know if you have pointers on the above ☝️

kasperisager:
Closing this in favour of MTU adaption directly in libudx.

PR #96, opened 2022-07-29
added test case for destroy nodes without ending socket

LuKks:
`node test/connections.js`

### DIFF
--- a/test/connections.js
+++ b/test/connections.js
@@ -3,6 +3,34 @@ const dgram = require('dgram')
 const { swarm } = require('./helpers')
 const DHT = require('../')
 
+test('createServer + connect - destroy nodes without ending socket', { timeout: 60000 }, async function (t) {
+  const [a, b] = await swarm(t)
+  const lc = t.test('socket lifecycle')
+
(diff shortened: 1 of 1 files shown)


test/connections.js
```diff
@@ -3,6 +3,34 @@ const dgram = require('dgram')
...
 
+test('createServer + connect - destroy nodes without ending socket', { timeout: 60000 }, async function (t) {
+  const [a, b] = await swarm(t)
+  const lc = t.test('socket lifecycle')
+
+  t.plan(2)
+  lc.plan(2)
+
+  const server = a.createServer(function (socket) {
+    lc.pass('server side opened')
```
mafintosh, line 14: you have to error handle the socket here, so it doesn't crash (ie attach an error listener that does nothing)

test/connections.js
```diff
@@ -3,6 +3,34 @@ const dgram = require('dgram')
...
+  })
+
+  await server.listen()
+
+  const socket = b.connect(server.publicKey)
+
```
mafintosh, line 20: same here

PR #97, opened 2022-07-30
Added a few test cases dealing with bootstraps

LuKks:
<s>The first test "bootstrapper" is actually a `dht-rpc` issue which I already opened a PR there, so it could be removed from here, no problem.</s>

I left some comments but before merging it would be appropriate to remove them.
Also, before merging I could edit the helper to be able to create those tests as the other ones.

### DIFF
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -3,28 +3,38 @@ const test = require('brittle')
 const DHT = require('../')
 // const { BOOTSTRAP_NODES } = require('../lib/constants.js')
 
-test('bootstrapper', async function (t) {
-  const node = DHT.bootstrapper(49737)
-
-  await node.ready()
-  t.is(typeof node.address().host, 'string')
-  t.is(typeof node.address().family, 'number')
-  t.is(typeof node.address().port, 'number')
-
-  await node.destroy()
-})
-
-test('local bootstrap with default node settings', async function (t) {
-  const bootstrap1 = new DHT({ bootstrap: [], port: 49737, anyPort: false, ephemeral: false, firewalled: false })
+test('local bootstrap with non ephemeral server', async function (t) {
+  const bootstrap1 = DHT.bootstrapper(49737, '127.0.0.1')
   await bootstrap1.ready()
 
-  const bootstrap = [bootstrap1.address()]
-  // const bootstrap = BOOTSTRAP_NODES // if you use default mainnet bootstraps, the test would work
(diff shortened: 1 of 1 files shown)


test/bootstrap.js
```diff
@@ -0,0 +1,105 @@
...
+  t.is(typeof node.address().port, 'number')
+
+  await node.destroy()
+})
+
+test('local bootstrap with default node settings', async function (t) {
+  const bootstrap1 = new DHT({ bootstrap: [], port: 49737, anyPort: false, ephemeral: false, firewalled: false })
+  await bootstrap1.ready()
+
+  const bootstrap = [bootstrap1.address()]
```
mafintosh, line 21: don't do this, as address is 0.0.0.0, do { port: bootstrap1.address().port, host: '127.0.0.1' }

test/bootstrap.js
```diff
@@ -0,0 +1,105 @@
...
+  await lc
+
+  await server.close()
+  await b.destroy()
+  await a.destroy()
+  await bootstrap1.destroy()
+})
+
+test('online bootstrap with default node settings', async function (t) {
+  const bootstrap = [{ host: 'dht1.lukks.ar', port: 49737 }]
```
mafintosh, line 68: let's not add any remote deps

test/bootstrap.js
LuKks: Yeap, this was already outdated as I learned it today with `dht-rpc`

test/bootstrap.js
LuKks: Right, it was temporal to show that it works with online bootstrap nodes. But right now my bootstrap doesn't seem to work at all, I created a new DigitalOcean server, etc but same, not working.

LuKks:
The 1st test `local bootstrap with non ephemeral server` works.
The 2nd test `local bootstrap with ephemeral server` doesn't work.
Why?

Also, the others commented out tests are working and they're basically the same as the two above except that are using online bootstrap nodes. Obviously no intention to merge them and they will be removed before do so.

LuKks:
@mafintosh Can you try to set up a functional bootstrap in the next server? 206.81.13.180 It's fresh from DigitalOcean

LuKks:
In case you don't trust dht1.lukks.ar and you want to do the remote bootstrap yourself, this is the new fresh server: 142.93.13.125 where you have SSH access and it's pointing to dht2.lukks.ar in case that's useful

LuKks:
Sorry my bad, it's all working perfectly. And by now the bootstrapper tests is already in `dht-rpc` so no need for this PR.

PR #98, opened 2022-07-30
Improving bootstrapping readme, examples, etc

LuKks:
<s>About the README, I understand that two is a better recommendation for redundancy, but just one seems also good enough to start.

Maybe this PR could wait until #97 is resolved to be able to use the `bootstrapper()` and remove the `ephemeral` setting, so that the example is simpler.</s>


LuKks:
Or maybe this way

```js
const bootstrap1 = DHT.bootstrapper(49737, '127.0.0.1')
await bootstrap1.ready()

const bootstrap = [{ host: '127.0.0.1', port: bootstrap1.address().port }]
```

mafintosh (commented):
Just use two bootstraps instead of a non-eph one. In practice you wanna use 2-3

mafintosh:
Also just notice think it should say isolated rather than private here in general, as private is an overloaded term

LuKks:
I think I could convert the explanation to a separated `.md` so if anyone wants to create their own DHT network (deal with bootstraps, persistent nodes, etc) could just read that `.md` as a way of tutorial. The explanation is mostly in chronological order so it makes more sense to read and understand.

LuKks:
I removed the `ephemeral` option from the basic-connectivity files to discourage the usage of ephemeral at all. It's true that they're demo files but it maybe would be a better example to follow if it's not there.

LuKks:
Totally hyperbolic, i.e. they could think: Oh, so my program is not a demo file so I will set it false.

PR #111, opened 2022-12-05
Destroy open streams on `dht.destroy()`

kasperisager:


kasperisager:
@mafintosh This is ready for review.

lib/raw-stream-set.js
```diff
@@ -26,4 +26,8 @@ module.exports = class RawStreamSet {
       self._streams.delete(id)
     }
   }
+
+  async destroy () {
+    for (const stream of this._streams.values()) stream.destroy()
```
kasperisager, line 31: Not sure if this should maybe wait for the streams to emit `close`?

mafintosh: I think it should

kasperisager: @mafintosh Should be all set now.

kasperisager:
@mafintosh This is ready for another look.

mafintosh:
Looks good. Test failure is due to a test not error handling i think (cause we are tearing down the streams aggresively now on test end)

LuKks:
Great, it also fixes https://github.com/hyperswarm/dht/issues/86
Only difference with v5 and v6 is that now it's giving `ECONNRESET` but that is correct as the socket was open

test/connections.js
```diff
@@ -408,6 +408,7 @@ test('relayed connection', async function (t) {
   const socket = b.connect(server.address().publicKey)
 
   server.on('connection', (socket) => {
+    socket.on('error', () => {})
```
LuKks, line 411: Sorry, have a question, why would this test give error? Both sides are gracefully doing `socket.end()`

kasperisager: The client side doesn't and is now forcefully destroyed on teardown.

mafintosh:
6.3.5

PR #115, opened 2022-12-21
CLI: add --port and --persistent for nodes

LuKks:


README.md
```diff
@@ -266,16 +266,53 @@ Any additional options you pass are forwarded to dht-rpc.
...
+import DHT from '@hyperswarm/dht'
+
+const node = new DHT({ bootstrap: ['server-ip:49737'] })
+```
+
+Note: You could configure some DNS for the bootstrap IP addresses.
+
+For the network to be fully operational it needs at least one persistent node.
+
+2) Provide the first persistent node using your own bootstrap values:
```
LuKks, line 307: If we ever fix the bootstrap node being able to act as a normal persistent node then we can get rid of this step!

LuKks: That complicates the setup when someone wants to create their own DHT network

PR #118, opened 2023-01-02
Remove udx-native and _udx prop, and return server on listen()

LuKks:


mafintosh:
(Its actually not removing udx, just using the one required in dht-rpc)

PR #121, opened 2023-01-04
Update README.md

LuKks:


README.md
```diff
@@ -189,7 +187,7 @@ The public key of the remote peer.
 
 #### `socket.publicKey`
 
-The connections public key.
+The public key of the server.
```
mafintosh, line 190: This is infact the pub key of the connection itself

LuKks: Is that just english semantics? Otherwise check this:

```js
server.on('connection', function (socket) {
  // All same:
  console.log(socket.publicKey.toString('hex'))
  console.log(keyPair.publicKey.toString('hex'))
  console.log(server.publicKey.toString('hex'))

  // Both true:
  console.log(socket.publicKey === keyPair.publicKey)
  console.log(server.publicKey === socket.publicKey)
})
```

README.md
```diff
@@ -31,11 +31,9 @@ const server = node.createServer()
 
 server.on('connection', function (socket) {
   // socket is E2E encrypted between you and the other peer
-  // pipe it somewhere like any duplex stream
-
   console.log('Remote public key', socket.remotePublicKey)
-  console.log('Local public key', socket.publicKey) // same as keyPair.publicKey
```
LuKks, line 37: If there is no need to emphasize `socket.publicKey` I would say we can remove it as it's not very common to use it (at least not when first trying it out)

README.md
```diff
@@ -189,7 +187,7 @@ The public key of the remote peer.
 
 #### `socket.publicKey`
 
-The connections public key.
+The public key of the local socket.
```
LuKks, line 190: What about this? Just above there is "remotePublicKey", so at the end it looks like this:

#### `socket.remotePublicKey`
The public key of the remote peer.

#### `socket.publicKey`
The public key of the local socket.

PR #124, opened 2023-01-14
Improve CLI

LuKks:
Don't merge yet. We first need https://github.com/mafintosh/dht-rpc/pull/58 and https://github.com/holepunchto/hyperswarm-dht/pull/117 to later update this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,22 +298,22 @@ Important: it requires the port to be open.
 
 Now your bootstrap node is ready to use at `(server-ip):49737`, for example:
 ```js
-const node = new DHT({ bootstrap: ['(server-ip):49737'] })
+const dht = new DHT({ bootstrap: ['(server-ip):49737'] })
 ```
 
 Note: You could configure some DNS for the bootstrap IP addresses.
 
 For the network to be fully operational it needs at least one persistent node.
 
-2) Provide the first persistent node using your own bootstrap values:
+2) Provide the first node using your own bootstrap values:
 
 ```sh
-hyperdht --port 49738 --bootstrap (server-ip):49737 --persistent
+hyperdht --port 49738 --bootstrap (server-ip):49737
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -269,15 +269,56 @@ See [dht-rpc](https://github.com/mafintosh/dht-rpc) for the additional APIs the
...
+
+2) Provide the first persistent node using your own bootstrap values:
+
+```sh
+hyperswarm-dht --port 49738 --bootstrap (server-ip):49737 --persistent
+```
+
+Important: it requires the port to be open like the bootstrap node.
+
+Don't use the `--persistent` flag again, or you could damage the network health.
```
mafintosh, line 318: lets not have the flag, footgun

LuKks:
Btw this also fixes this case where incorrectly creates a bootstrap node:
`hyperswarm-dht --bootstrap <host:port> --host <ip> --port <n>`

README.md
LuKks: 100%, just depends on https://github.com/mafintosh/dht-rpc/pull/58

LuKks:
I still left the `--persistent` in the CLI but removed from the docs, it's actually useful when you're playing with own DHT networks but can't be recommended to use. Ideally the bootstrap node should serve for routing peers as per https://github.com/mafintosh/dht-rpc/pull/58

bin.js
```diff
@@ -21,32 +23,55 @@ function arg (name) {
...
+    })
+
+    node.on('close', function () {
+      console.log('Node #' + id + ' closed')
     })
   }
 
   console.log('Fully started ' + cnt + ' Hyperswarm DHT node' + (cnt === 1 ? '' : 's'))
+
+  process.once('SIGINT', function () {
```
mafintosh, line 72: you should print here, "Shutting down nodes..."

PR #125, opened 2023-02-03
Holepunching on local networks

LuKks:


lib/server.js
```diff
@@ -254,6 +255,19 @@ module.exports = class Server extends EventEmitter {
       return hs
     }
 
+    if (this._shareLocalAddress && !direct && clientAddress.host === serverAddress.host) {
```
LuKks, line 258: This added condition of `_shareLocalAddress` fixes two failing tests, but just realized it's wrong because server might still want to do local connections or at least it's still capable of doing so, even if it's not sharing local addresses

lib/server.js
```diff
@@ -254,6 +255,19 @@ module.exports = class Server extends EventEmitter {
...
+      const clientAddresses = remotePayload.addresses4.filter(onlyPrivateHosts)
+
+      if (clientAddresses.length > 0) {
```
LuKks, line 261: Replying the previous review, it doesn't seem necessary to guard against serverSocket not existing

test/holepuncher.js
```diff
@@ -0,0 +1,148 @@
...
+
+  t.alike(Holepuncher.matchAddress(
+    [{ host: '192.168.122.238' }],
+    [{ host: '192.168.0.23' }, { host: '192.168.122.1' }, { host: '172.16.1.1' }]
+  ), { host: '192.168.122.1' }, 'Same subnet (third segment)')
+
+  /* t.alike(Holepuncher.matchAddress(
+    [{ host: '192.168.122.238' }],
+    [{ host: '192.168.0.23' }, { host: '192.168.122.1' }, { host: '192.168.122.238' }, { host: '172.16.1.1' }]
+  ), { host: '192.168.122.238' }, 'Full match') */
```
mafintosh, line 147: why is this commented out?

LuKks: It doesn't work because I invented that list of addresses, so I think it's invalid that it has this two:
`{ host: '192.168.122.1' }, { host: '192.168.122.238' }`

Currently it will pick up `192.168.122.1`, but I think only "routers" have the `.1`, right? For example, if my computer has a docker network, and a virtual machines network, then my same computer will have two networks that ends with `.1`

mafintosh: makes sense that it picks the first one

LuKks:
@kasperisager Ready for review!

lib/holepuncher.js
```diff
@@ -297,4 +305,55 @@ function coerceFirewall (fw) {
...
+          if (best.segment === 2) best = { segment: 3, addr: remoteAddress }
+
+          // Matches 192.168.122.*
+          if (a[2] === b[2]) {
+            if (best.segment === 3) best = { segment: 4, addr: remoteAddress }
+          }
+        }
+      }
+    }
+  }
```
kasperisager, line 354: There should never be more addresses than would be feasible for a quadratic search to look through, no?

LuKks: Very true, before there was but code has changed

PR #129, opened 2023-03-11
Use parent DHT host

LuKks:
@riccardobl proposed this some time ago! This time I used the host from `dht.io`

Reference: https://github.com/riccardobl/hyperswarm-dht/commit/df9aa0d81bc88b08163ce263580e6603305ab44e

### DIFF
--- a/lib/socket-pool.js
+++ b/lib/socket-pool.js
@@ -117,7 +117,7 @@ class SocketRef {
       .on('message', this._onmessage.bind(this))
       .on('idle', this._onidle.bind(this))
       .on('busy', this._onbusy.bind(this))
-      .bind()
+      .bind(0, pool._dht.io._host)
 
     this._refs = 1
     this._released = false


LuKks:
This doesn't necessarily depends on https://github.com/mafintosh/dht-rpc/pull/63 but they go together

LuKks:
@mafintosh Ready!

lib/socket-pool.js
```diff
@@ -117,7 +117,7 @@ class SocketRef {
       .on('message', this._onmessage.bind(this))
       .on('idle', this._onidle.bind(this))
       .on('busy', this._onbusy.bind(this))
-      .bind()
+      .bind(0, pool._dht.io._host)
```
mafintosh, line 120: this is accessing a private prop. feauture is okay, but then dht-rpc should expose something like dht.listening (null or the host its bound to) etc (naming tdb)

mafintosh: Actually thinking this through, this doesn't make sense currently, as it needs much more information that just the "parent host". The swarm runs (not fully yet but will) in dual mode, meaning it uses ipv6 and ipv4 and dht-rpc only runs in the ipv4 so this turns massively complex really fast.

The solution WHEN we enable dual mode is that hyperswarm-dht has options for choosing what interfaces it prefers for both and passes them back up to dht-rpc, ie the other way

mafintosh:
Reverse flow of this implemented in https://github.com/holepunchto/hyperswarm-dht/commit/ebef0db6024f5fba0be6e7c9c344624681e5ca06 for now

PR #135, opened 2023-08-29
Implement blind relay support

kasperisager:
This PR implements client connection support for the blind relay protocol (https://github.com/holepunchto/protomux-blind-relay). This allows peers running in both client and server mode to establish connections to other peers through blind relays. Additionally, an initial implementation of a connection pool has been added, providing a way to more efficiently manage and deduplicate connections to other peers. ~~When provided with a connection pool, peers will use the pool for outbound connections to blind relays.~~


lib/connection-pool.js
```diff
@@ -0,0 +1,146 @@
...
+  get connecting () {
+    return this._connecting.size
+  }
+
+  get connections () {
+    return this._connections.values()
+  }
+
+  has (publicKey) {
+    const keyString = publicKey.toString('hex')
```
mafintosh, line 112: b4a

lib/connection-pool.js
```diff
@@ -0,0 +1,146 @@
...
+
+    return this._connections.has(keyString) || this._connecting.has(keyString)
+  }
+
+  get (publicKey) {
+    const keyString = publicKey.toString('hex')
```
mafintosh, line 118: b4a

lib/connection-pool.js
```diff
@@ -0,0 +1,146 @@
...
+    super()
+
+    this._dht = dht
+    this._servers = new Map()
+    this._connecting = new Map()
+    this._connections = new Map()
+  }
+
+  _attachServer (server) {
+    const keyString = server.publicKey.toString('hex')
```
mafintosh, line 16: b4a

lib/connect.js
```diff
@@ -357,6 +385,48 @@ async function connectThroughNode (c, address, socket) {
     c.rawStream = null
   }
 
+  if (payload.relayThrough || c.relayThrough) {
+    let isInitiator, publicKey, token
```
mafintosh, line 389: make 3 lets

kasperisager:
Removed use of the connection pool for outbound connections to relays for now and instead destroy the relay socket when the raw stream closes. This isn't optimal, but should at least be correct.

lib/connect.js
```diff
@@ -294,7 +311,11 @@ async function connectThroughNode (c, address, socket) {
         id: c.rawStream.id,
         seq: 0
       },
-      secretStream: {}
+      secretStream: {},
+      relayThrough: c.relayThrough && {
```
mafintosh, line 315: prefer a ternary here that falls back to null

lib/server.js
```diff
@@ -259,6 +285,52 @@ module.exports = class Server extends EventEmitter {
...
+        token = remotePayload.relayThrough.token
+      }
+
+      hs.relayToken = token
+      hs.relaySocket = this.dht.connect(publicKey)
+      hs.relayClient = relay.Client.from(hs.relaySocket, { id: hs.relaySocket.publicKey })
+
+      hs.relayClient
+        .pair(isInitiator, token, hs.rawStream)
+        .on('error', () => hs.relaySocket.destroy())
```
mafintosh, line 309: you don't need to attach on(close) for this? (and ignore onerror) - it might close without an error but unsure what the flow is here

kasperisager: The pairing request will emit an error if the pairing is cancelled, in which case we can immediately destroy the relay socket. Otherwise, if the pairing succeeds, we maintain the relay socket until the raw stream closes

mafintosh (approved):

package.json
```diff
@@ -26,15 +27,17 @@
     "hypercore-crypto": "^3.3.0",
     "noise-curve-ed": "^2.0.0",
     "noise-handshake": "^3.0.0",
+    "protomux": "^3.5.1",
```
mafintosh, line 30: dont think this is used?

package.json
```diff
@@ -26,15 +27,17 @@
...
     "record-cache": "^1.1.1",
     "safety-catch": "^1.0.1",
     "sodium-universal": "^4.0.0",
+    "udx-native": "^1.7.2",
```
mafintosh, line 34: do we require udx here?

kasperisager: No, only needed it as `udx-native` 1.7.2 wasn't tagged `latest`.

test/all.js
```diff
@@ -9,10 +9,13 @@ async function runTests () {
 
   await import('./announces.js')
   await import('./connections.js')
+  await import('./holepuncher.js')
```
mafintosh, line 12: ouch

lanmower:
Thank you both of you, this is huge. 🙏🥳

PR #144, opened 2023-10-06
Fix flaky tests

kasperisager:


lib/connect.js
```diff
@@ -100,12 +100,6 @@ module.exports = function connect (dht, publicKey, opts = {}) {
   }
 
   function firewall (socket, port, host) {
-    // Check if the traffic originated from the socket on which we're expecting relay traffic. If so,
-    // we haven't hole punched yet and the other side is just sending us traffic through the relay.
-    if (c.relaySocket && c.relaySocket.rawStream && c.relaySocket.rawStream.socket === socket) {
-      return false
-    }
```
kasperisager, line 107: @mafintosh This was causing the relay tests to sometimes stall, but removing it without another solution in place isn't really an option either.

I'm thinking a more robust solution would be to avoid the problem entirely by introducing a new protocol message that tells the other end of the relayed connection that they should start sending data. A peer can then safely assume that they will only receive data on the relayed connection _after_ they've connected their end of the stream to the relay socket.

kasperisager: Let me document the message flow though, that should also help visualise it.

kasperisager: https://github.com/holepunchto/blind-relay?tab=readme-ov-file#protocol

What I was thinking was that after each peer receives a "pair" message from the relay, they'd connect their raw streams but not yet start the encrypted streams. They'd then send a new message, "ready"/"connect"/..., to the relay which would forward it to the remote peer. Only upon receiving this message would a peer start the encrypted stream.

kasperisager:
@mafintosh This is ready for review.

PR #152, opened 2023-11-21
fix LAN handshake failure leak

mafintosh:


HDegroote (approved):
Deployed with this fix, and the leak was fixed (`swarm.server._holepunches` no longer grows arbitrarily large). The other behaviour seems stable, so this PR looks good to me

PR #154, opened 2023-12-18
Use catch(safetyCatch) instead of noop

HDegroote:
Might need to be merged together with https://github.com/holepunchto/hyperswarm/pull/170, because hyperswarm at some point removes a `noop` listener created in hyperdht's `lib/connect` (although I couldn't trace which one exactly)

### DIFF
--- a/lib/connection-pool.js
+++ b/lib/connection-pool.js
@@ -1,6 +1,7 @@
 const EventEmitter = require('events')
 const b4a = require('b4a')
 const errors = require('./errors')
+const safetyCatch = require('safety-catch')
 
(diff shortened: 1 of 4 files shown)


lib/connection-pool.js
```diff
@@ -40,24 +41,24 @@ module.exports = class ConnectionPool extends EventEmitter {
...
           })
           .destroy(errors.DUPLICATE_CONNECTION())
 
         stream
-          .on('error', noop)
+          .on('error', safetyCatch)
           .on('close', onclose)
       } else {
         stream
-          .on('error', noop)
```
mafintosh, line 60: all of these should be noops

PR #157, opened 2024-01-03
close, resume, suspend waits for listen

mafintosh:


lib/server.js
```diff
@@ -54,12 +58,14 @@ module.exports = class Server extends EventEmitter {
     this.emit('connection', encryptedSocket)
   }
 
-  suspend () {
+  async suspend () {
```
HDegroote, line 61: hyperdht's index.js was already assuming `suspend` and `resume` were async, so this is fine if these methods were internal-only. If not, we should check that they're awaited in other modules calling `server.suspend()` and `server.resume()`

HDegroote: Oops, missed they were implicitly async already

PR #158, opened 2024-01-04
 Add hypertrace

freeall:
Adding hypertrace instrumentation. Also added to `dht-rpc`.


mafintosh:
we did this in a sep pr

PR #169, opened 2024-03-20
Add relay test (catches bug for udx-native <1.8.9)

freeall:


test/integration/fixtures/server-through-relay.js
```diff
@@ -0,0 +1,37 @@
...
+  }, socket => {
+    socket.setKeepAlive(socketKeepAlive)
+    socket
+      .on('data', data => {
+        console.log(`socket_ondata ${data.toString()}`)
+        socket.write('world')
+      })
+      .on('open', () => console.log(`socket_onopen ${socket.rawStream.remoteHost}:${socket.rawStream.remotePort}`))
+      .on('close', () => console.log('socket_onclose'))
+      .on('error', err => console.log(`socket_onerror ${err.code}`))
```
HDegroote, line 28: Detail: I don't think `err` is guaranteed to have a `code` prop, so `err?.code` seems safer (same for the other err handler)

HDegroote: Edit: it's actually fine, since it will just print null, so nvm

test/integration/keep-alive-with-relay.js
```diff
@@ -0,0 +1,140 @@
...
+  let didClientNotDetectThatServerDiedTimer
+  let didClientNotDetectThatServerDiedTimerFired = false
+
+  t.teardown(async () => {
+    await clientNode.destroy()
+    await relayNode.destroy()
+  })
+
+  await startRelayServer()
+  startServer()
```
HDegroote, line 46: async function, so I think it needs awaiting or a catch-handler to avoid uncaught exceptions

test/integration/keep-alive-with-relay.js
```diff
@@ -0,0 +1,140 @@
...
+        // when error is ETIMEDOUT it's the server connection that has broken
+        // not so long after that, the client should have detected that the connection is gone
+        if (err.code === 'ETIMEDOUT') {
+          relayTest.pass('Relay server detected that server has died. Waiting for client to detect')
+
+          // In some cases, the client may have detected that the server has died before the relay server
+          if (hasClientDetectedThatServerDied) return
+
+          const timeToDetectClientHasDied = SOCKET_KEEPALIVE
+          didClientNotDetectThatServerDiedTimer = setTimeout(() => {
```
HDegroote, line 69: Detail: probably could do with some shorter names for these variables if it's possible to still have them be clear

test/integration/keep-alive-with-relay.js
```diff
@@ -0,0 +1,140 @@
...
+
+  const RELAY_KEEPALIVE = 500
+  const SOCKET_KEEPALIVE = 10 * RELAY_KEEPALIVE
+
+  const clientKeyPair = DHT.keyPair()
+  const clientNode = new DHT()
+  const relayNode = new DHT()
+  const relayKeyPair = DHT.keyPair()
+  const serverKeyPair = DHT.keyPair()
+  const serverPublicKey = serverKeyPair.publicKey.toString('hex')
```
HDegroote, line 34: Detail: same as with `Buffer.from`, the `b4a` way is `b4a.toString(serverKeyPair.publicKey, 'hex')`. But I'm not sure if we plan to update the tests to use b4a throughout

package.json
```diff
@@ -44,6 +44,8 @@
   },
   "devDependencies": {
     "brittle": "^3.0.0",
+    "graceful-goodbye": "^1.3.0",
+    "keypear": "^1.1.1",
```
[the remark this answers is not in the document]
freeall: Oh yeah, you're right. I think it was a rebase that went wrong.

test/integration/keep-alive-with-relay.js
freeall: Don't agree on this. I think the long names - especially - test avoid misunderstandings. It just looks silly :)

HDegroote: Okay, makes sense to me to keep them then

HDegroote (approved):
Looks good to me

PR #177, opened 2024-04-02
add connectionKeepAlive opt

HDegroote:
TODO's before marking ready:
- Should the keepAlives for the relays use the passed-in value if non-0, or remain fixed at 5s? (https://github.com/holepunchto/hyperdht/blob/79eea93b25b975715a769d2f3440994224fab7d3/lib/server.js#L613 and https://github.com/holepunchto/hyperdht/blob/79eea93b25b975715a769d2f3440994224fab7d3/lib/connect.js#L693)
- Cut a hyperswarm-secret-stream release with https://github.com/holepunchto/hyperswarm-secret-stream/pull/30 so the tests don't rely on a private prop


mafintosh:
6.14.0

PR #167, opened 2024-03-18
add invalidatesStream to handshake

mafintosh:

PR #183, opened 2024-06-26
add timestamp to handshake

billiegoose:
Supercedes #167 . It's better because:
- it invalidates old streams without requiring any old state, meaning if the program crashes and restarts, it'll invalidate any old streams as long as `timestamp` is  monotonically increasing ¹ ²
- it's also simpler and requires less book-keeping code in `hyperswarm`
- a timestamp is shorter than a noise stream id
- a timestamp is the same length as a udx stream id, but doesn't suffer from relays changing the udx stream id

¹  `Date.now()` is _not_ guaranteed to be monotonically increasing, but since fast reconnects are not mission critical, it should suffice, since it'll succeed "all" the time for practical purposes.

² According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Performance/timeOrigin), `performance.timeOrigin + performance.now()` _is_ guaranteed to be monotonically increasing. But I don't think the `performance` global is available in `bare` JS environments.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -414,8 +414,7 @@ async function connectThroughNode (c, address, socket) {
     serverAddress,
     payload
   }
-  c.encryptedSocket.userData ??= {}
-  c.encryptedSocket.userData.timestamp = payload.timestamp
+  c.encryptedSocket.timestamp = payload.timestamp
 
   c.payload = new SecurePayload(hs.holepunchSecret)
 
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -81,7 +81,10 @@ module.exports = function connect (dht, publicKey, opts = {}) {
     relaySocket: null,
     relayClient: null,
     relayPaired: false,
-    relayKeepAlive: opts.relayKeepAlive || 5000
+    relayKeepAlive: opts.relayKeepAlive || 5000,
+
+    // Used by the server when tie-breaking between two connections from the same peer
+    timestamp: opts.timestamp ?? -1
   }
 
   // If the raw stream receives an error signal pre connect (ie from the firewall hook), make sure
@@ -374,7 +377,8 @@ async function connectThroughNode (c, address, socket) {
       secretStream: {},
       relayThrough: c.relayThrough
         ? { publicKey: c.relayThrough, token: c.relayToken }
-        : null
+        : null,
+      timestamp: c.timestamp
     })
     if (isDone(c)) return
   }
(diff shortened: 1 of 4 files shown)


lib/connect.js
```diff
@@ -81,7 +81,10 @@ module.exports = function connect (dht, publicKey, opts = {}) {
     relaySocket: null,
     relayClient: null,
     relayPaired: false,
-    relayKeepAlive: opts.relayKeepAlive || 5000
+    relayKeepAlive: opts.relayKeepAlive || 5000,
+
+    // Used by the server when tie-breaking between two connections from the same peer
+    timestamp: opts.timestamp ?? -1
```
billiegoose, line 87: this default of `-1` makes the timestamp behavior _opt in_... which may be a pain-in-the-butt for most users, who'll not benefit from this change until they modify their code to pass `{ timestamp: Date.now() }` in the options to `connect`.

However, I tend to be very conservative when it comes to breaking changes, and my philosophy is usually to make all enhancements "opt in" until the next major version bump, at which time I update the defaults.

lib/connect.js
```diff
@@ -410,6 +414,8 @@ async function connectThroughNode (c, address, socket) {
     serverAddress,
     payload
   }
+  c.encryptedSocket.userData ??= {}
+  c.encryptedSocket.userData.timestamp = payload.timestamp
```
billiegoose, line 418: is using `userData` ugly? idk. The comment for it in secret-stream says "// pointer for upstream to set data here if they want". Are we upstream? Should we modify secret-stream so it has a timestamp property proper?

lib/connect.js
mafintosh: just use 0 as the default, then the uint can be unsigned, and 0 is pretty appropriate for that. also means we can just do || 0

lib/messages.js
```diff
@@ -162,6 +162,7 @@ exports.noisePayload = {
     if (m.udx) udxInfo.preencode(state, m.udx)
     if (m.secretStream) secretStreamInfo.preencode(state, m.secretStream)
     if (m.relayThrough) relayThroughInfo.preencode(state, m.relayThrough)
+    if (m.timestamp >= 0) c.uint.preencode(state, m.timestamp)
```
mafintosh, line 165: just > 0

billiegoose: IDK, being able to send a 0 timestamp could be useful... 😆

billiegoose:
```suggestion
    if (m.timestamp > 0) c.uint.preencode(state, m.timestamp)
```

billiegoose:
Closed in favor of https://github.com/holepunchto/hyperswarm/pull/178

PR #184, opened 2024-07-02
add a 'relaying' event to connections

billiegoose:
https://app.asana.com/0/1206774237110658/1207704548548702

Intent - there should be a `.relay` property on connection objects that accurately reflects the relaying state of the connection. And it should fire events when that state changes.

```js
swarm.on('connection', (conn) => {
  function cb () {
    console.log('currently using relay?', conn.relay.relaying)
  }
  conn.relay.on('relay', cb)
  conn.relay.on('unrelay', cb)
}
```

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -421,7 +421,6 @@ async function connectThroughNode (c, address, socket) {
 
     if (c.rawStream.connected) {
       c.rawStream.once('remote-changed', () => {
-        c.dht.tracer.trace('remote-changed', { relay: c.encryptedSocket.relay })
         this.encryptedSocket.emit('unrelay', c.encryptedSocket.relay)
       })
       const remoteChanging = c.rawStream.changeRemote(socket, c.connect.payload.udx.id, port, host)
@@ -740,7 +739,6 @@ function relayConnection (c, relayThrough, payload, hs) {
   }
 
   function onabort (err) {
-    c.dht.tracer.trace('relay-aborted', { relayThrough: publicKey, remotePublicKey: c.remotePublicKey })
     if (c.relayTimeout) clearRelayTimeout(c)
     const socket = c.relaySocket
     c.relayToken = null
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -29,6 +29,7 @@ const {
   RELAY_ABORTED,
   SUSPENDED
 } = require('./errors')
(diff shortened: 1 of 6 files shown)


lib/connect.js
```diff
@@ -690,6 +690,7 @@ function relayConnection (c, relayThrough, payload, hs) {
     publicKey = relayThrough
     token = c.relayToken
   }
+  c.encryptedSocket.emit('relaying', { relayThrough: publicKey })
```
mafintosh, line 693: just emit(relaying, publicKey) is fine

mafintosh: ah nvm, it pairs nicely with the other one, would add remotePublicKey: c.remotePublicKey also here so its the same object emitted always between the conn and the server

lib/server.js
```diff
@@ -613,6 +613,8 @@ module.exports = class Server extends EventEmitter {
       token = remotePayload.relayThrough.token
     }
 
+    this.emit('relaying', { publicKey: h.remotePublicKey, relayThrough: publicKey })
```
mafintosh, line 616: remotePublicKey:

lib/connect.js
```diff
@@ -1,3 +1,5 @@
+/** @import HyperDHT from '../index' */
+/** @import UDXStream from 'udx-native/lib/stream' */
```
billiegoose, line 2: I find these annotations immensely helpful, because it enables language-server based IDEs to link everything. I'm happy to remove them from this PR, or to fully annotate the holepunch ecosystem so IDEs grok it.

mafintosh: lets keep that separate, already complex stuff here

test/relaying.js
```diff
@@ -420,6 +444,17 @@ test('relay connections through node, client and server side', async function (t
...
+  bSocket.relay
+    .on('relay', () => {
+      lc.pass('client socket is being relayed')
+      lc.is(bSocket.relay.relaying, true)
+      lc.alike(bSocket.relay.remotePublicKey, bServer.publicKey)
+      lc.alike(bSocket.relay.relayThrough, aServer.publicKey)
+    })
+    .on('abort', () => {
+      lc.fail('client relay aborted')
+    })
```
billiegoose, line 456: AFAICT this is the only `relaying.js` test that actually relays. All other tests abort the relay.

PR #186, opened 2024-07-26
connectionKeepAlive: 5000 default

HDegroote:
When connectionKeepAlive is not set, strange things happen, particularly if only one side has not set it. Behaviour will be cleaner if it is set by default.

Note: as this change rolls out, we might temporarily create more situations where one side has set the keepAlive, while the other side hasn't.


index.js
```diff
@@ -29,7 +29,9 @@ class HyperDHT extends DHT {
     this.defaultKeyPair = opts.keyPair || createKeyPair(opts.seed)
     this.listening = new Set()
     this.tracer = createTracer(this)
-    this.connectionKeepAlive = opts.connectionKeepAlive || 0
+    this.connectionKeepAlive = opts.connectionKeepAlive == null
```
HDegroote, line 32: This still supports passing in a 0 value. We can also prohibit a 0 value by doing `this.connectionKeepAlive = opts.connectionKeepAlive || 5000`

test/integration/keep-alive.js
```diff
@@ -62,7 +62,7 @@ test('Client not using keepalive does not detect disconnect - separated by proce
 
   const clientTest = t.test('client')
   const { bootstrap } = await swarm(t)
-  const node = new DHT({ bootstrap })
+  const node = new DHT({ bootstrap, connectionKeepAlive: 0 })
```
HDegroote, line 65: This test tests the case where no keepAlive is set (previously default, now explicit). We might just remove this test now, as it's not that relevant anymore

index.js
HDegroote: Cleaned up the contract to `opts.connectionKeepAlive === false ? 0 : opts.connectionKeepAlive || 5000`

PR #187, opened 2024-07-29
Unslab remotePublicKey in connect

HDegroote:
Currently causes a memleak in hyperswarm-secret-stream (its `publicKey` retains ~8kb buffers, its `remotePublicKey` retains ~68kb buffers)


HDegroote:
There's no point in unslabbing the publicKey, since it's our own key we use everywhere (so only one buffer)

PR #193, opened 2024-09-24
Populate DHT with Pear.config.dht

AndreiRegiani:
* Renamed `HyperDHT.addNode()` (very misleading name, this actually checks for skipping testnet nodes) renamed to `filterNode()`

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -19,7 +19,7 @@ exports.BOOTSTRAP_NODES = [
   '138.68.147.8@node3.hyperdht.org:49737'
 ]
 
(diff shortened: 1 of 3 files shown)


lib/constants.js
```diff
@@ -19,6 +19,8 @@ exports.BOOTSTRAP_NODES = [
   '138.68.147.8@node3.hyperdht.org:49737'
 ]
 
+exports.KNOWN_NODES = global?.Pear?.config?.dht || []
```
mafintosh, line 22: doing global.Pear?.config.dht should be fine, global is always set and config is always set if Pear is

PR #208, opened 2024-11-20
Throw error when reusing keyPair for listens

HDegroote:


lib/server.js
```diff
@@ -153,8 +153,14 @@ module.exports = class Server extends EventEmitter {
     await this.dht.bind()
     if (this._closing) return
 
-    this.target = unslabbedHash(keyPair.publicKey)
+    for (const s of this.dht.listening) {
+      if (s._keyPair && b4a.equals(s._keyPair.publicKey, keyPair.publicKey)) {
+        this.dht.listening.delete(this)
```
HDegroote, line 158: I don't think this is needed, but left it in to sanity check.

(If it is needed, then we need to try-finally `await this.dht.bind()`, because that can throw too)

HDegroote: We delete when `server._gc` is called (when `server.close()`, or when `await this._announcer.start()` errors).

But if there's an error while calling server.listen(), the server will remain in `this.dht.listening`, unless the user explicitly closes it.

The current approach is definitely bad. I think we either leave it to the user (as when `await this.dht.bind rejects) or pass through gc, as when the announcer errors (https://github.com/holepunchto/hyperdht/blob/b5cd15a31f75dc27a0657593baebd2680fae6090/lib/server.js#L171-L177)

The cleanest might be to wrap everything from `await this.dht.bind()` up to `await this._announcer.start()` in the try-finally

mafintosh: Thats a feature atm. The user owns the server until they call close on it, ever or not listen fails.

HDegroote: Aha, in that case I'll simply remove the delete and it should be good

HDegroote: I had misunderstood. I pushed a commit to modify the behaviour so it always cleans up if _listen throws before announcer.start() completes

PR #236, opened 2026-02-13
Add holepunching support on LAN

supersuryaansh:
Currently on LAN after discovering that both peers are on LAN, the client just pings the server using a dht.ping() and waits for it to respond, however if the server is behind a firewall that ping never resolves and we throw a HOLEPUNCH_ABORTED error where we never actually holepunch.

This PR changes that and enables holepunching b/w LAN clients as well.

Info: I used AI to help me with this PR and I have written it to the best of my knowledge.


marcus-pousette-hp:
Fature added in 6.33.0!

PR #241, opened 2026-03-17
Add PHT node put/get for search stack

noahlevenson:
**EDIT, APRIL 10 2026: this PR has been tentatively deprecated in favor of the approach in https://github.com/holepunchto/hyperdht/pull/246**

Today, HyperDHT powers Hyperswarm.  Tomorrow, it will also power Hypersearch.  This PR implements primitives for Hypersearch.

More specifically:

This PR implements a put/get interface for prefix hash tree nodes.  The PHT nodes are authenticated in two ways:  cryptographically, in that the DHT target for a PHT node is derived from the publisher's public key -- and structurally, in that a PHT node must conform to the expected schema.

A useful prefix hash tree is a linked data structure consisting of a set of PHT nodes.  Thus, the intention of the put/get interface implemented in this PR is to enable any DHT participant to define and publish static PHT topologies which cannot be tampered with by other participants.  By "static," we mean immutable tries which cannot add or delete keys. (This sort of trie is useful if you know all your keys ahead of time, as is the case with a static keyword vocabulary.)

Note that "authenticated" does not mean "immutable."  Bob can publish a trie consisting of a set of PHT nodes -- and then later, there's nothing stopping him from overwriting it by publishing new nodes using the same public key and tree ID.  But this would be a case of user error, and Bob should not do this.


lejeunerenard:
@noahlevenson with #246 merged, this can be closed correct?

noahlevenson:
@lejeunerenard Correct, we can close it.

PR #244, opened 2026-03-26
Safety catch the probes

HDegroote:
I'm not sure it's guaranteed that they never throw, so might as well be conservative


lejeunerenard:
Closing as the methods should never throw and if they do, then catching would be masking a bug.

PR #245, opened 2026-04-06
Pre-connect to `closestNodes` in parallel to `FIND_PEER` query

lejeunerenard:
Builds on the `nodes-in-key` PR #237 to use `closestNodes` from either the key, cache or passed option and attempts connecting through the `closestNodes` in parallel to querying the DHT for more fresh nodes. This should skip the query when the cached node is still good and will speed up connections especially on slower devices where querying the DHT is more expensive because of latency or dropped packets.

The `closestNodes` are still passed to the `FIND_PEER` to speed up the query in the case that the closestNodes have updated their record of the `target`. The normal 2 tries is kept.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -379,11 +379,7 @@ async function findAndConnect(c, opts) {
         }
 
         // Skip node already run via preConnect
-        if (
-          preConnect &&
-          closestNodes &&
-          closestNodes.find((n) => n.host === data.from.host && n.port === data.from.port)
-        ) {
+        if (preConnect && closestNodes && isClosestNode(closestNodes, data)) {
           sem.signal()
           continue
         }
@@ -877,6 +873,14 @@ function selectRelay(relayThrough) {
   if (Array.isArray(relayThrough))
     return relayThrough[Math.floor(Math.random() * relayThrough.length)]
   return relayThrough
+}
+
+function isClosestNode(closestNodes, data) {
+  for (const node of closestNodes) {
+    if (node.host === data.from.host && node.port === data.from.port) return true
+  }
+
+  return false
 }
 
 function noop() {}
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -381,6 +381,7 @@ async function findAndConnect(c, opts) {
         // Skip node already run via preConnect
         if (
           preConnect &&
+          closestNodes &&
           closestNodes.find((n) => n.host === data.from.host && n.port === data.from.port)
         ) {
           sem.signal()
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -334,30 +334,31 @@ async function connectThroughNodes(c, addresses, socket) {
 
 async function findAndConnect(c, opts) {
   let attempts = 0
-  let closestNodes = opts.relayAddresses && opts.relayAddresses.length ? opts.relayAddresses : null
+  let relayAddresses =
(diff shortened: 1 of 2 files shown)


lib/connect.js
```diff
@@ -364,6 +378,16 @@ async function findAndConnect(c, opts) {
           break
         }
 
+        // Skip node already run via preConnect
+        if (
+          preConnect &&
+          closestNodes &&
+          closestNodes.find((n) => n.host === data.from.host && n.port === data.from.port)
```
mafintosh: make a static function `isClosestNode(closestNodes, data)` or similar and do a simple loop in that

lib/connect.js
```diff
@@ -364,6 +378,16 @@ async function findAndConnect(c, opts) {
           break
         }
 
+        // Skip node already run via preConnect
+        if (
+          preConnect &&
+          closestNodes &&
```
mafintosh: if preconnect is set closestNodes is set afaik

lib/connect.js
```diff
@@ -341,18 +350,23 @@ async function findAndConnect(c, opts) {
   }
 
   // 2 is how many parallel connect attempts we want to do, we can make this configurable
-  const sem = new Semaphore(2)
+  const sem = new Semaphore(closestNodes !== null ? 3 : 2)
```
Drache93: Do you want the length check here also? Otherwise we'll be doing 3 even if cloestNodes empty

lib/connect.js
```diff
@@ -341,18 +350,23 @@ async function findAndConnect(c, opts) {
...
 
   try {
     for (let i = 0; i < tries && !isDone(c) && !c.connect; i++) {
       c.query = c.dht.findPeer(c.target, {
         hash: false,
         session: c.session,
         closestNodes,
-        onlyClosestNodes: closestNodes !== null,
-        retries: closestNodes ? 1 : 3
+        retries: 3
```
Drache93, line 370: Is this less about the preConnect changes. And more the general insight we found that retries === 1 is just too little?

lib/connect.js
Drache93: Also does the length check when setting preconnect

Drache93 (commented):
Really nice

lib/connect.js
lejeunerenard: They are set together only initially. `closestNodes` will be nulled after the first try here:
https://github.com/holepunchto/hyperdht/pull/245/changes#diff-459561ac30417a9e40a7904e9a3c297a6c840189bf21583d951746a1d43a506aL372
I actually hit this when testing across the stack. The second try is incase the closestNodes are not valid and a fully new query needs to be done (much like the current implementation).

lib/connect.js
lejeunerenard: completed here: 5f115e51dba78c00ff97cc79a4d4c90141d74fe0

lib/connect.js
lejeunerenard: This is because the `1` retry count only applied when it only queried closest nodes. Now that closest nodes are only used to help get new nodes here (aka we query vs trying to immediately connect) and `onlyClosestNodes` is always false, it can treat it as a general query that we don't need to bail early from if it fails.

lib/connect.js
lejeunerenard: great point. i'll adjust

lejeunerenard: fixed here: 5738f01b98b6b628458493314e184e3673da7f76

PR #246, opened 2026-04-10
Plugin system

noahlevenson:
Implements a system for injecting new persistent commands as fully encapsulated plugins. 

Aspiring plugin authors must simply extend the `Plugin` base class, implement `onrequest` and `destroy`, and `register` their new plugin with HyperDHT.

See tests and forthcoming Hypersearch plugin for examples.


test/plugins.js
```diff
@@ -0,0 +1,173 @@
+const test = require('brittle')
+const c = require('compact-encoding')
+const sodium = require('sodium-universal')
+const HyperDHT = require('../')
```
lejeunerenard:
```suggestion
```

Never used

test/plugins.js
```diff
@@ -0,0 +1,173 @@
...
+      if (!req.value) return
+
+      let plugreq
+      try {
+        plugreq = c.decode(m.pluginRequest, req.value)
+      } catch {
+        return
+      }
+
+      const { plugin, command, payload } = plugreq
```
lejeunerenard:
```suggestion
      const { command, payload } = plugreq
```
`plugin` isn't used

test/plugins.js
```diff
@@ -0,0 +1,173 @@
...
+    destroy() {
+      // Do nothing
+    }
+
+    onput(req) {
+      if (!req.target || !req.token || !req.value) return
+
+      let val
+      try {
+        const { plugin, command, payload } = c.decode(m.pluginRequest, req.value)
```
lejeunerenard:
```suggestion
        const { payload } = c.decode(m.pluginRequest, req.value)
```

lib/messages.js
```diff
@@ -419,3 +419,23 @@ exports.mutableGetResponse = {
...
+  encode(state, m) {
+    c.string.encode(state, m.plugin)
+    c.int.encode(state, m.command)
+    c.buffer.encode(state, m.payload)
+  },
+  decode(state) {
+    return {
+      plugin: c.string.decode(state),
+      command: c.int.decode(state),
+      payload: c.buffer.decode(state)
```
lejeunerenard: Nit: My immediate thought was that `payload` should be `value` to match the top level command structure. Theoretically conveying that a plugin command is another type of command with a prefix of string.

test/plugins.js
```diff
@@ -0,0 +1,173 @@
...
+  class TestPlugin extends DHTPlugin {
+    constructor(dht) {
+      super('testplugin')
+
+      this.dht = dht
+      this.data = new Map()
+    }
+
+    onrequest(req) {
+      if (!req.value) return
```
lejeunerenard: This is unnecessary because of the check in `Persistent.onplugin` right?

lib/persistent.js
```diff
@@ -227,6 +227,20 @@ module.exports = class Persistent {
...
 
+  onplugin(req) {
+    if (!req.value) return
+
+    const plugreq = decode(m.pluginRequest, req.value)
+    if (plugreq === null) return
+
+    const { plugin, command, payload } = plugreq
+
+    const p = this.dht.plugins.get(plugin)
```
lejeunerenard:
```suggestion
    const p = this.dht.plugins.get(plugreq.plugin)
```

lib/constants.js
```diff
@@ -10,7 +10,8 @@ const COMMANDS = (exports.COMMANDS = {
   MUTABLE_PUT: 6,
   MUTABLE_GET: 7,
   IMMUTABLE_PUT: 8,
-  IMMUTABLE_GET: 9
+  IMMUTABLE_GET: 9,
+  PLUGIN_PERSISTENT: 10
```
lejeunerenard: Why specify `PERSISTENT`? I'd vote it be `PLUGIN`.

lib/plugin.js
```diff
@@ -0,0 +1,13 @@
+module.exports = class Plugin {
+  constructor(name) {
+    this.name = name
+  }
```
lejeunerenard, line 13: We should include a method to encode a request & query so the plugin author doesn't need to know the internals.

Something like:
```js
  request({ token = null, command, target = null, payload = null }, to, opts) {
    const value = c.encode(m.pluginRequest, {
      plugin: this.name,
      command,
      payload
    })

    return this.dht.request({
      token,
      target,
      command: HYPERDHT_COMMANDS.PLUGIN_PERSISTENT,
      value
    }, to, opts)
  }

  query({ command, target = null, payload = null }, opts) {
    const value = c.encode(m.pluginRequest, {
      plugin: this.name,
      command,
      payload
    })

    return this.dht.query({
      target,
      command: HYPERDHT_COMMANDS.PLUGIN_PERSISTENT,
      value
    }, opts)
  }
```

otherwise the plugin will repeatedly be (en/de)coding the plugin request.

lib/plugin.js
```diff
@@ -0,0 +1,13 @@
...
+
```
lejeunerenard, line 14: Should also have a `onregister(dht)`. A simple default implementation with subclasses implementing anything more sophisiticated:
```js
  onregister(dht) {
    this.dht = dht
  }
```

index.js
```diff
@@ -509,6 +515,10 @@ class HyperDHT extends DHT {
       from
     )
   }
+
+  register(name, plugin) {
+    this.plugins.set(name, plugin)
```
lejeunerenard, line 520:
```suggestion
    this.plugins.set(name, plugin)
    plugin.onregister(this)
```
To support the `onregister()` hook here: https://github.com/holepunchto/hyperdht/pull/246/changes#r3081222517

Needs further changes if implementing the plugin as a property in the original design.

test/plugins.js
```diff
@@ -0,0 +1,173 @@
...
+
+  function mapTest(node) {
+    return node
+  }
+
+  class TestPlugin extends DHTPlugin {
+    constructor(dht) {
+      super('testplugin')
+
+      this.dht = dht
```
lejeunerenard:
```suggestion
    constructor() {
      super('testplugin')
```
Can be simplified if we use the `onregister` hook here:
https://github.com/holepunchto/hyperdht/pull/246/changes#r3081222517

lib/persistent.js
```diff
@@ -227,6 +227,20 @@ module.exports = class Persistent {
...
+    if (!p) return
+
+    p.onrequest(req)
```
lejeunerenard, line 239: Should the "request" we send to the plugin be scoped down to the plugin command specifics? This would not put it on the plugin to decode the plugin value themselves and would match the `onrequest` callback for the top level comands.

Something like:
```js
    p.onrequest({
      ...req,
      reply: req.reply.bind(req),
      error: req.error.bind(req),
      // TODO maybe implement? sendReply: req.sendReply.bind(req),
      ...plugreq
    })
```

index.js
```diff
@@ -126,6 +127,7 @@ class HyperDHT extends DHT {
     }
     this._router.destroy()
     if (this._persistent) this._persistent.destroy()
+    for (const [_, plugin] of this.plugins) plugin.destroy()
```
lejeunerenard:
```suggestion
    for (const plugin of this.plugins.values()) plugin.destroy()
```

test/plugins.js
```diff
@@ -0,0 +1,173 @@
...
+    const p = new TestPlugin(node)
+    pluginClients.push(p)
+    node.register(p.name, p)
+  }
+
+  const put = await pluginClients[30].put('myTestValue')
+
+  t.is(put.target.length, 32)
+
+  const res = await pluginClients[30].get(put.target)
```
lejeunerenard, line 140: Originally the plugin would defined on `node.testplugin` aka (`node[pluginname]`). It feels a bit error prone in that the plugin name needs to be a valid property identifier and can't conflict with existing properties.

But the alternative of managing the plugin instances like above seems also not desirable. Maybe a compromise is that you get the plugin via `node.plugins.get('testplugin')` instead of the direct property originally proposed?

For reference, here is the original property based implementation I sketched while reviewing:
https://github.com/holepunchto/hyperdht/commit/f01a64aed5248882d25c185ad48887a8a2082ea5

lejeunerenard (commented):
Looks great! I have some nits, suggestions and discussion points as comments.

The tl;dr is:
- we should encapsulate `request` & `query` for the plugins by default so plugin request (en/de)coding isn't necessary
- further towards the objective of making plugin commands look like top level commands, maybe we should rename `payload` to `value`?
- we should support a `onregister` hook
- we should add a `onpersistent` hook [example](https://github.com/holepunchto/hyperdht/commit/4575e671a049599992f8ca771f187acf60b9b0f9)
- plugins being accessible via properties vs manual management vs using `dht.plugins.get(name)`

I made a branch showing most of my suggestions changes and their impacts to plugin implementation as a reference here:
https://github.com/holepunchto/hyperdht/tree/ljr/plugins
We can merge, discard, etc as needed, but thought seeing the changes would help discussions around them.

noahlevenson:
Cool, all makes sense to me.  I'm adding a couple questions/reactions to your comments mostly for my own education.  I'm happy to merge your suggestions branch.

lib/messages.js
noahlevenson: I don't really like `payload` either.  I think the counterargument to `value` is that there's an added layer of indirection at the plugin command layer -- that is, our plugin commands  are essentially multiplexed over a single top level command, and so there's another layer that gets unwrapped -- so it might be worth disambiguating it. I don't feel too strongly about it, though.

lejeunerenard: Yeah agree. That was nice in the example branch to dissembiguate, but I also accidentally mix them up too because I absent mindedly assumed they were the same.

It can go either way, but I vote to use `value` for both given the intent for plugin commands to essentially be the same as top level commands in functionality. That gives an edge to that option IHO.

lib/constants.js
noahlevenson: I don't have an opinion about this, but FWIW, here was my rationale as a noob to the HyperDHT codebase:

Reading the code and observing the formalization of ephemeral vs. persistent states was kind of confusing.  AFAIK, the terms are not common in the DHT literature.  I can grok what they mean here and the purpose they serve in HyperDHT, but I ask myself:  why are they named this way -- is it because there's other context that I'm missing?

When I looked at the paths in `hyperdht.onrequest`, I asked myself:  would it be so crazy to write an ephemeral plugin that implements an alternate handshake / holepunching?  I thought:  no, that doesn't seem crazy.  

So I figured the plugin commands should accommodate that.

test/plugins.js
noahlevenson: I defer to HyperDHT veterans on such questions of ergonomics.  To my sensibility, though, I'm happy enough with plugins in a `Map`.  Are you specifically referring to the practice of keeping an array of plugin clients, parallel to the DHT swarm, for testing?  That pattern could go away if the plugin shipped with test helpers like HyperDHT does.  Right?

noahlevenson:
Actually, wait -- there's one thing I don't understand:  adding the `onpersistent` hook.

Since you must register a plugin with HyperDHT at runtime, sometime after it's constructed -- and since HyperDHT bootstraps (and tries to go persistent) at construction time -- I wonder:  why do we have both an `onpersistent` hook **and** runtime registration?  If we want the `onpersistent` hook, shouldn't the API require you to provide all your plugins as `opts` at construction time?

I feel like I'm probably misunderstanding something?

test/plugins.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/011cc8af88d3b89d8bc3a15410e88244efb49c1a

test/plugins.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/011cc8af88d3b89d8bc3a15410e88244efb49c1a

test/plugins.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/011cc8af88d3b89d8bc3a15410e88244efb49c1a

lib/messages.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/011cc8af88d3b89d8bc3a15410e88244efb49c1a

test/plugins.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/011cc8af88d3b89d8bc3a15410e88244efb49c1a

lib/persistent.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/011cc8af88d3b89d8bc3a15410e88244efb49c1a

index.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/f05d10d3b532bdb6f5d2e336c2017c768bc2a71b

lib/plugin.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/2f02a6bc688907a1bd66a0c7f46221c32a3e5adc

index.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/2f02a6bc688907a1bd66a0c7f46221c32a3e5adc

test/plugins.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/6fd8e3afe5843f2c30299cf9039b1ddc6bbc7313

lib/plugin.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/2f02a6bc688907a1bd66a0c7f46221c32a3e5adc

lib/persistent.js
noahlevenson: If I'm understanding this correctly, there's an aspect of it that I'm not sure about:

We've renamed `pluginRequest.payload` to `pluginRequest.value`, so here we're creating a new `req` that has the properties from the original `req` but with the `value` and `command` replaced by the value and command extracted from the encapsulated plugin request.  Right?

I personally find this too subtle and implicit, particularly for a new plugin system that's still evolving.  I think it obfuscates the plugin decapsulation in a way that's hard to debug.

Philosophically, I tend to think that protocol layering (and decapsulation) should always be explicit. 

So I don't really want a magic `req` that's the hidden merger of two protocol layers -- I want to write decapsulation boilerplate.  That's just me, though.

lejeunerenard:
> Actually, wait -- there's one thing I don't understand: adding the `onpersistent` hook.
> 
> Since you must register a plugin with HyperDHT at runtime, sometime after it's constructed -- and since HyperDHT bootstraps (and tries to go persistent) at construction time -- I wonder: why do we have both an `onpersistent` hook **and** runtime registration? If we want the `onpersistent` hook, shouldn't the API require you to provide all your plugins as `opts` at construction time?
> 
> I feel like I'm probably misunderstanding something?

That's a great question. I think `.register()` as the registration method, while done at runtime, should be synchronously (aka in the same tick & not a microtask) after the construction of a node. I think of this as the same as Express.js's `.use()` so if you register a plugin later, then there is no way for the node to process requests that happen before nor the `onpersistent` hook. Theoretically if a plugin needed to register later, but needed to know if the node was persistent, it could via the private property `node._persistent`. Not that we want to encourage that.

I'm in favor of keeping the `.register()` pattern for the added flexibility with the above caveat.

lib/constants.js
lejeunerenard: I think the command enum doesn't need to assert whether it's persistent or not as I assume the same command enum would be used for any scenario as it is a way to partition off commands for plugins from the rest of the DHT commands.

If we decide later that ephemeral nodes should be able to respond to plugin commands we can refactor then moving the handler above the check and out of `Persistent`. Nothing that I can see depends on the `Persistent` class such that we end up locking ourselves in. For now until we find a use case otherwise, requiring a persistent node aligns with how the DHT is used generally. Persistent nodes are reliable vs ephemeral nodes so can respond to commands.

test/plugins.js
lejeunerenard: Not specifically referring to keeping them in an array, no. Just requiring the user of the plugin to manage it. The arrays are as you reference more of a testing requirement.

I think a map is best as it avoids conflicts with properties on the node itself.

lib/persistent.js
lejeunerenard: For the `value` name I think it still makes sense as 'value' is what you want to send with the command.


For the mixing of the top level `req` with the plugin `onrequest`'s `req`, yeah you're right. It's too magical and mixes the two layers of abstraction too much. Maybe the plugin's `req` can have property for the wrapping request? The using any info or methods from the wrapping request would be an explicit choice for the plugin author. I definitely don't think we should require plugins to decode the top level `value` that encodes the plugin's request. But I can see that methods like `reply()` would be convenient.

lib/constants.js
noahlevenson: Cool, will rename it.

FWIW:

> I assume the same command enum would be used for any scenario as it is a way to partition off commands for plugins from the rest of the DHT commands.

I think this is the part that I disagree with.  If a plugin could be ephemeral or persistent, I think that having a single `PLUGIN` command in the enum creates more complexity.  AFAICT -- If you wrote a persistent plugin, but the handler fires in the ephemeral state, you have to duplicate the persistent check in `hyperdht.onrequest` in your plugin's `onrequest` method.

lejeunerenard: The command comes from the sender and so they can't be certain that you'd be in either state anyways. The node needs to check itself if it is persistent or not [the same as it does currently](https://github.com/holepunchto/hyperdht/blob/3de277c2a03d8978f348a7dfcfe98728665ca052/index.js#L403) for anything other than the holepunching commands, right?

noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/30ec8aded1b58b1f846316e6ce45461efe5441cb

lib/persistent.js
noahlevenson: IMO, we don't want two `onrequest(req)` interfaces that are both exactly the same except the `req` is subtly different in one of the places.  I think that kind of stuff drives engineers crazy.  

But I think it's challenging to identify the optimal abstraction right now, since we only have one use case (hypersearch).  

My rationale was thus to leave it very simple and unoptimized for now, and to expose the full unadulterated `req` to the plugin.  

I'd wager that the next few use cases will likely help reveal what the function signature for `plugin.onrequest` should be.  (Maybe the right solution is actually to rename `plugin.onrequest` to something else, to help convey that it's actually a different layer of abstraction and receives different inputs).  

But I find it challenging to reason about any proposed modification, because it just feels like premature abstraction to me?

lib/messages.js
```diff
@@ -419,3 +419,23 @@ exports.mutableGetResponse = {
...
+    c.buffer.encode(state, m.value)
+  },
+  decode(state) {
+    return {
+      plugin: c.string.decode(state),
+      command: c.int.decode(state),
+      value: c.buffer.decode(state)
+    }
+  }
+}
```
lejeunerenard, line 444:
```suggestion
exports.pluginRequest = {
  preencode(state, m) {
    c.string.preencode(state, m.plugin)
    c.uint.preencode(state, m.command)
    c.buffer.preencode(state, m.value)
  },
  encode(state, m) {
    c.string.encode(state, m.plugin)
    c.uint.encode(state, m.command)
    c.buffer.encode(state, m.value)
  },
  decode(state) {
    return {
      plugin: c.string.decode(state),
      command: c.uint.decode(state),
      value: c.buffer.decode(state)
    }
  }
}
```
Command field should be `uint`.

lejeunerenard (changes requested):

lib/messages.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/commits/89d874d6e4c4194c0b99ed9f1dfee458cb0b4b70

lib/persistent.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/changes/73999f37168ec358b4fe1d432ebda3e3a211cc10

test/plugins.js
lejeunerenard: A plugin's `onrequest()` doesn't need to return a bool correct?

test/plugins.js
```diff
@@ -54,26 +45,20 @@ test('plugin put - get', async function (t) {
       // Do nothing
     }
 
-    onput(req) {
-      if (!req.target || !req.token || !req.value) return
+    onput(req, outerReq) {
+      if (!outerReq.target || !outerReq.token || !outerReq.value) return
```
lejeunerenard: `outerReq.value` doesn't need to be checked because it's checked in `onplugin()` in the persistent class.

test/plugins.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/commits/5439ed1c4a21debeb6dfa450c9ff42563f4ed093

test/plugins.js
noahlevenson: https://github.com/holepunchto/hyperdht/pull/246/commits/5439ed1c4a21debeb6dfa450c9ff42563f4ed093

index.js
```diff
@@ -509,6 +516,11 @@ class HyperDHT extends DHT {
       from
     )
   }
+
+  register(name, plugin) {
+    this.plugins.set(name, plugin)
```
lejeunerenard, line 521:
```suggestion
  register(plugin) {
    this.plugins.set(plugin.name, plugin)
```
`.register()` should probably only take the plugin instance, right? Since the plugin's `.name` is used to lookup the plugin on a plugin request, it should register using `.name` directly instead of a dev passed `name`?

noahlevenson: Depends on whether you use the plugin convenience methods, `plugin.request` and `plugin.query`.

If you don't use the convenience methods, you can send a request invoking a command on a plugin that's been registered under an alternate name.  Right?

lejeunerenard: That's true, but then the `.name` of the plugin wouldn't be used.

noahlevenson: Yep, that's correct. 

Making `register()` an arity-2 function instead of an arity-1 function  just provides a little escape hatch for testing, prototyping, development, etc. You can register a forked plugin under an alias and have it running alongside the mainline plugin.

Since it costs nothing, I'm not sure why we'd get rid of it?

lejeunerenard: Great point. Discussion a bit with @andrewosh out of band and it does sound useful to have forked plugins simultaneously. We'll keep the 2 arity version.

PR #248, opened 2026-04-15
Fix relay abort propagation

marcus-pousette-hp:
`updateHolepunch` was previously called with wrong arguments, and bad test coverage did allow this behaviour previously

Tests where co-written with AI

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -726,7 +726,7 @@ function maybeDestroyEncryptedSocket(c, err) {
 
 async function abort(c, { peerAddress, relayAddress }, err) {
   try {
-    await updateHolepunch(peerAddress, relayAddress, {
+    await updateHolepunch(c, peerAddress, relayAddress, {
       error: ERROR.ABORTED,
       firewall: FIREWALL.UNKNOWN,
       round: c.round++,
--- a/test/relaying.js
+++ b/test/relaying.js
@@ -128,19 +128,7 @@ test('relay connections through node, server side, client abort notifies remote'
 
   const lc = t.test('socket lifecycle')
   lc.plan(5)
-
-  let holepunchCalls = 0
-  let holepunchCallsAtAbort = -1
-  const peerHolepunch = c._router.peerHolepunch.bind(c._router)
-
-  c._router.peerHolepunch = async function (...args) {
-    holepunchCalls++
-    return peerHolepunch(...args)
-  }
-
-  t.teardown(() => {
-    c._router.peerHolepunch = peerHolepunch
-  })
+  let clientAbortedUpgrade = false
 
   const relay = new RelayServer({
     createStream(opts) {
@@ -177,11 +165,13 @@ test('relay connections through node, server side, client abort notifies remote'
 
   await bServer.listen()
 
+  const remoteAbort = waitFor(() => bServer._holepunches.some((hs) => hs && hs.aborted))
+
   const bSocket = c.connect(bServer.publicKey, {
     fastOpen: false,
     localConnection: false,
     holepunch() {
-      holepunchCallsAtAbort = holepunchCalls
+      clientAbortedUpgrade = true
       return false
     }
   })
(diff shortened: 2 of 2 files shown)


lib/connect.js
```diff
@@ -755,7 +726,7 @@ function maybeDestroyEncryptedSocket(c, err) {
 
 async function abort(c, { peerAddress, relayAddress }, err) {
   try {
-    await updateHolepunch(peerAddress, relayAddress, {
+    await updateHolepunch(c, peerAddress, relayAddress, {
```
HDegroote, line 729: Nice catch (a bit embarrassing). This threw as a TypeError, right? IMO we should replace the catch-all handler with a safety-catch handler, which would have bubbled this up much sooner

test/relaying.js
```diff
@@ -119,6 +119,96 @@ test('relay connections through node, client side, client aborts hole punch', as
...
+  let holepunchCallsAtAbort = -1
+  const peerHolepunch = c._router.peerHolepunch.bind(c._router)
+
+  c._router.peerHolepunch = async function (...args) {
+    holepunchCalls++
+    return peerHolepunch(...args)
+  }
+
+  t.teardown(() => {
+    c._router.peerHolepunch = peerHolepunch
```
HDegroote: Why is this needed?

test/relaying.js
```diff
@@ -119,6 +119,96 @@ test('relay connections through node, client side, client aborts hole punch', as
...
+    .on('open', () => {
+      lc.pass('client socket opened')
+    })
+    .on('close', () => {
+      lc.pass('client socket closed')
+    })
+    .end('hello world')
+
+  await lc
+  await new Promise((resolve) => setTimeout(resolve, 50))
```
HDegroote: what are you waiting for exactly here? If it's linearisable, that makes for a cleaner test. If not, better to put something like 200ms to compensate for slow CI machines

marcus-pousette-hp: This is a bit smelly here. Yes. Right now it's delay-based, but what I actually care about is the extra peerHolepunch() call after the client aborts the upgrade path.  A proper fixed would be event based. I will try to do that

test/relaying.js
marcus-pousette-hp: Not perfectly necessary. Monkey-patching the c local method does not create any leak problems if test exits normally and instances are cleared up.  I will remove it.

HDegroote: There's no way at all it can leak right? Because it's monkey-patching the instance's method, not the method for the class (unless I'm misunderstanding the code).

if it's actually monkey-patching HyperDHT itself then it's a risky test

marcus-pousette-hp: Yeap exactly

test/relaying.js
```diff
@@ -119,6 +119,95 @@ test('relay connections through node, client side, client aborts hole punch', as
...
+    .on('close', () => {
+      lc.pass('client socket closed')
+    })
+    .end('hello world')
+
+  await lc
+  await remoteAbort
+
+  t.ok(clientAbortedUpgrade, 'client reached the holepunch decision point')
+  t.pass('remote records the client abort')
```
HDegroote, line 194: What's the use of this assertion?

test/relaying.js
```diff
@@ -119,6 +119,95 @@ test('relay connections through node, client side, client aborts hole punch', as
...
+  const b = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
+  const c = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
+
+  const lc = t.test('socket lifecycle')
+  lc.plan(5)
+  let clientAbortedUpgrade = false
+
+  const relay = new RelayServer({
+    createStream(opts) {
+      return a.createRawStream({ ...opts, framed: true })
```
HDegroote, line 139: (detail) Worth adding another `lc.pass('sanity check: using the relay')` assertion here to sanity check it's actually using the relay

HDegroote (approved):
Looks good to me (just some minor test comments, but nothing blocking merge--just to make the PR cleaner). Up to Sean to merge

test/relaying.js
marcus-pousette-hp: Good catch, no use anymore/redundant since the assertion is done differently now

marcus-pousette-hp: I will fix

test/relaying.js
marcus-pousette-hp: Great catch again. I will add it

lib/connect.js
marcus-pousette-hp: Yes! 

https://github.com/holepunchto/hyperdht/pull/250 here is a follow up PR for two places where we would benefit by using safety-catch

lejeunerenard (approved):

PR #249, opened 2026-04-15
Test/relay direct upgrade coverage

marcus-pousette-hp:
Adds missing test coverage for the upgrade path: 

that the relayed connection is first -> and then later upgrades to direct

and that this happens *in place*  on the same raw stream

PR co-written with AI

### DIFF
--- a/test/relaying.js
+++ b/test/relaying.js
@@ -437,8 +437,10 @@ test('relay connection upgrades to direct connection', { timeout: 30000 }, async
   const b = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
   const c = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
 
+  // Capture raw-stream transitions so we can prove the socket upgrades in place.
   instrumentRawStreams(t, b)
   instrumentRawStreams(t, c)
+  // Give the relay path a short head start so the upgrade ordering is deterministic.
   delayPunching(t, 500)
 
   const relay = new RelayServer({
@@ -491,10 +493,12 @@ test('relay connection upgrades to direct connection', { timeout: 30000 }, async
   t.is(clientSocket.rawStream._transitions[0].type, 'connect')
   t.is(serverSocket.rawStream._transitions[0].type, 'connect')
 
+  // The relayed connection should already be usable before the direct path wins.
   const beforeUpgrade = once(clientSocket, 'data')
   clientSocket.write(Buffer.from('before upgrade'))
   t.alike((await beforeUpgrade)[0], Buffer.from('before upgrade'), 'relay path carries data')
 
+  // changeRemote is the signal that the same raw stream switched transport.
   await waitForUpgrade(clientSocket.rawStream)
   await waitForUpgrade(serverSocket.rawStream)
 
@@ -603,6 +607,7 @@ function instrumentRawStreams(t, dht) {
     const stream = createRawStream(...args)
     stream._transitions = []
 
+    // Record the initial relay connect and the later direct changeRemote.
     const connect = stream.connect.bind(stream)
     stream.connect = function (socket, id, port, host) {
       stream._transitions.push({ type: 'connect', host, port })
@@ -627,6 +632,7 @@ function delayPunching(t, ms) {
   const punch = Holepuncher.prototype._punch
 
   Holepuncher.prototype._punch = async function () {
+    // Bias the race without changing the real upgrade logic.
     await new Promise((resolve) => setTimeout(resolve, ms))
     return punch.call(this)
   }
@@ -639,6 +645,7 @@ function delayPunching(t, ms) {
 async function waitForUpgrade(stream) {
   const started = Date.now()
 
+  // Poll until the raw stream reports its remote endpoint changed.
   while (!hasUpgrade(stream)) {
     if (Date.now() - started > 5000) {
       throw new Error('Timed out waiting for relay-to-direct upgrade')
--- a/test/relaying.js
+++ b/test/relaying.js
@@ -437,11 +437,7 @@ test('relay connection upgrades to direct connection', { timeout: 30000 }, async
   const serverNode = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
   const clientNode = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
 
-  // Capture raw-stream transitions so we can prove the socket upgrades in place.
-  instrumentRawStreams(t, serverNode)
-  instrumentRawStreams(t, clientNode)
-  // Bias the race so relay wins first for these nodes only; upgrade assertions below are still condition-based.
-  delayPunching(t, 500, [serverNode, clientNode])
+  const resumePunching = pausePunching(t, [serverNode, clientNode])
 
   const relayServer = new RelayServer({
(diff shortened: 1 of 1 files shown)


test/relaying.js
```diff
@@ -428,6 +430,106 @@ test('relay connections through node, client and server side', async function (t
...
+  t.teardown(() => relayServer.close())
+
+  const relayTransportServer = relayNode.createServer(function (socket) {
+    const session = relayServer.accept(socket, { id: socket.remotePublicKey })
+    session.on('error', (err) => t.comment(err.message))
+  })
+
+  await relayTransportServer.listen()
+
+  let serverSocket = null
```
lejeunerenard: nit: `serverSocket` could be gotten from the `serverSocketOpened` promise since that resolves with the socket as well.

test/relaying.js
```diff
@@ -428,6 +430,106 @@ test('relay connections through node, client and server side', async function (t
...
+
+  await appServer.listen()
+
+  const clientSocket = clientNode.connect(appServer.publicKey, {
+    fastOpen: false,
+    localConnection: false
+  })
+
+  await once(clientSocket, 'open')
+  await serverSocketOpened
```
lejeunerenard: These can be awaited together, correct?

test/relaying.js
```diff
@@ -428,6 +430,106 @@ test('relay connections through node, client and server side', async function (t
...
+  t.is(serverSocket.rawStream._transitions[0].type, 'connect')
+
+  // The relayed connection should already be usable before the direct path wins.
+  const beforeUpgrade = once(clientSocket, 'data')
+  clientSocket.write(Buffer.from('before upgrade'))
+  t.alike((await beforeUpgrade)[0], Buffer.from('before upgrade'), 'relay path carries data')
+
+  // changeRemote is the signal that the same raw stream switched transport.
+  await waitForUpgrade(clientSocket.rawStream)
+  await waitForUpgrade(serverSocket.rawStream)
```
```suggestion
  await Promise.all([
    once(clientSocket.rawStream, 'remote-changed'),
    once(serverSocket.rawStream, 'remote-changed')
 ])
```

You can use the `remote-changed` event on UDX streams for this which is better than polling.  In this case they have to be awaited together given they fire at the same time.

test/relaying.js
```diff
@@ -498,6 +600,64 @@ test.skip('relay several connections through node with pool', async function (t)
...
+  // Poll until the raw stream reports its remote endpoint changed.
+  while (!stream._transitions.some((transition) => transition.type === 'changeRemote')) {
+    if (Date.now() - started > 5000) {
+      throw new Error('Timed out waiting for relay-to-direct upgrade')
+    }
+
+    await new Promise((resolve) => setTimeout(resolve, 20))
+  }
+}
+
```
```suggestion
```
Assuming `remote-changed` is used instead, this can be removed.

test/relaying.js
```diff
@@ -498,6 +600,64 @@ test.skip('relay several connections through node with pool', async function (t)
...
+function delayPunching(t, ms, delayedNodes) {
+  const punch = Holepuncher.prototype._punch
+
+  Holepuncher.prototype._punch = async function () {
+    // Only delay the punchers created by this test.
+    if (delayedNodes.includes(this.dht)) {
+      // Bias the race without changing the real upgrade logic.
+      await new Promise((resolve) => setTimeout(resolve, ms))
```
lejeunerenard: I think this can flake if the timing is off. You could return a function that calls the `resolve` here to enable holepunching for the `delayedNodes` manually in the test.

test/relaying.js
```diff
@@ -498,6 +600,64 @@ test.skip('relay several connections through node with pool', async function (t)
   await c.destroy()
 })
 
+function instrumentRawStreams(t, dht) {
```
lejeunerenard: Instead of instrumenting/monkey patching the stream, you can check the `remoteHost` & `remotePort` to see that they are the values expected.

test/relaying.js
```diff
@@ -428,6 +430,106 @@ test('relay connections through node, client and server side', async function (t
...
+    localConnection: false
+  })
+
+  await once(clientSocket, 'open')
+  await serverSocketOpened
+
+  t.is(clientSocket.rawStream._transitions.length, 1, 'client starts on the relayed stream')
+  t.is(serverSocket.rawStream._transitions.length, 1, 'server starts on the relayed stream')
+  t.is(clientSocket.rawStream._transitions[0].type, 'connect')
+  t.is(serverSocket.rawStream._transitions[0].type, 'connect')
```
lejeunerenard: These assertions only tell us that the client & server connected to something. I think we can rely on the write message below to do that for us.

We could instead check that the servers's rawStream `.remotePort` doesn't match the clients `.localPort`. Then check that they do match when direct connected. Using this approach we don't need to use any private properties or methods and don't need to monkey patch the streams.

lejeunerenard (changes requested):

test/relaying.js
marcus-pousette-hp: Ye! way better

marcus-pousette-hp: Resolved!

test/relaying.js
marcus-pousette-hp: Thanks!

test/relaying.js
marcus-pousette-hp: Thanks! this is way better

test/relaying.js
marcus-pousette-hp: Yeap!

test/relaying.js
marcus-pousette-hp: this is great!

test/relaying.js
marcus-pousette-hp: For sure this is way better, thanks

test/relaying.js
```diff
@@ -589,6 +596,31 @@ test.skip('relay several connections through node with pool', async function (t)
...
+  const punch = Holepuncher.prototype._punch
+  let resumed = false
+  let resume = null
+  const punchingResumed = new Promise((resolve) => {
+    resume = function () {
+      if (resumed) return
+      resumed = true
+      resolve()
+    }
+  })
```
lejeunerenard: No need to guard against double firing. The promise doesn't change state once resolved. So `resume` can be assigned to `resolve` directly.

marcus-pousette-hp: Yes makes sense, addressed!

test/relaying.js
```diff
@@ -519,6 +430,102 @@ test('relay connections through node, client and server side', async function (t
   await a.destroy()
 })
 
+test('relay connection upgrades to direct connection', { timeout: 30000 }, async function (t) {
```
```suggestion
test('relay connection upgrades to direct connection', async function (t) {
```
The timeout is not needed.

test/relaying.js
```diff
@@ -519,6 +430,102 @@ test('relay connections through node, client and server side', async function (t
...
+  const { bootstrap } = await swarm(t)
+
+  const relayNode = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
```
```suggestion
  const relayNode = createDHT({ bootstrap })
```

The relay can (and probably should) be persistent, right? The server & client can remain ephemeral to ensure they don't quick connect, but the relay can.

lejeunerenard (approved):

PR #250, opened 2026-04-16
Use safety-catch in cleanup paths

marcus-pousette-hp:
As as a consequence of #248 I introduce 2 places where we should do safetyCatch instead of swallowing all error in case there are errors like TypeError that should not be caught 

Waiting for #248 to be reviewed so we can make CI pass

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -117,7 +117,9 @@ module.exports = class Server extends EventEmitter {
 
     try {
       await this._listening
-    } catch {}
+    } catch (err) {
+      safetyCatch(err)
+    }
 
     this._gc()
     this._clearAll()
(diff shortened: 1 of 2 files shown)


lib/server.js
```diff
@@ -117,7 +117,9 @@ module.exports = class Server extends EventEmitter {
 
     try {
       await this._listening
-    } catch {}
+    } catch (err) {
```
HDegroote, line 120: Neutral about this one. Since we're closing here, I care less about swallowing potential errors--up to Sean. The other one definitely makes sense to me

lejeunerenard (approved):

lib/server.js
lejeunerenard: Also fairly neutral but none of the existing errors should be rethrown by `safetyCatch` so opting to include.

PR #251, opened 2026-04-28
Close relay sockets after direct upgrade

marcus-pousette-hp:
#249 verifies that relayed connections can upgrade to a direct path in place: the raw stream starts relayed, emits `remote-changed`, switches to the peer address, and continues carrying data.

This PR adds the missing cleanup behavior for that path. Since relay sockets are currently opened per relayed connection, once both sides upgrade to direct there is no active use for the relay transport sockets anymore. Keeping them open leaves extra HyperDHT connections and keepalives around for the lifetime of the app connection.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -483,6 +483,9 @@ async function connectThroughNode(c, address, socket) {
     if (c.rawStream === null) return // Already hole punched
 
     if (c.rawStream.connected) {
+      // Once this fires, changeRemote has moved the raw stream onto the direct path.
+      c.rawStream.once('remote-changed', () => closeRelayConnection(c))
+
       const remoteChanging = c.rawStream.changeRemote(socket, c.connect.payload.udx.id, port, host)
 
       if (remoteChanging) remoteChanging.catch(safetyCatch)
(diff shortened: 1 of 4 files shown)


lib/connect.js
```diff
@@ -806,21 +809,19 @@ function relayConnection(c, relayThrough, payload, hs) {
...
     c.relayPaired = true
 
-    const { remotePort, remoteHost, socket } = c.relaySocket.rawStream
+    // Keep a stable reference since c.relaySocket is cleared after direct upgrade.
+    const relaySocket = c.relaySocket
+    const { remotePort, remoteHost, socket } = relaySocket.rawStream
 
     c.rawStream
-      .on('close', () => c.relaySocket.destroy())
+      .on('close', () => relaySocket.destroy())
```
lejeunerenard: Use the new `destroyRelayConnection()` instead then the `relaySocket` ref doesn't need to be kept around until the direct connection closes.

lib/server.js
```diff
@@ -665,10 +668,12 @@ module.exports = class Server extends EventEmitter {
...
         hs.prepunching = null
 
-        const { remotePort, remoteHost, socket } = hs.relaySocket.rawStream
+        // Keep a stable reference since hs.relaySocket is cleared after direct upgrade.
+        const relaySocket = hs.relaySocket
+        const { remotePort, remoteHost, socket } = relaySocket.rawStream
 
         hs.rawStream
-          .on('close', () => hs.relaySocket.destroy())
+          .on('close', () => relaySocket.destroy())
```
lejeunerenard: Same here.

lib/server.js
```diff
@@ -694,6 +695,30 @@ function clearRelayTimeout(hs) {
...
+
+  const socket = hs.relaySocket
+  // Drop relay references so the app connection no longer keeps relay state alive.
+  hs.relayToken = null
+  hs.relaySocket = null
+  hs.relayClient = null
+
+  return socket
+}
+
```
lejeunerenard: We should unify client & server logic for managing the relay connection closing & timer since its the same. Moving these functions into a shared module that both uses maybe naming the arg `connectionOrHandshake` instead of `c` & `hs` respectively to show that it can be either. But all properties & methods are the same.

This should also reflects the future where relay connections are reused better since a relay could be reused once as a server and once as a client on the same peer.

lejeunerenard (changes requested):

lib/server.js
marcus-pousette-hp: Yeap, very much makes sense and reducing total LOC. Fixed!

lejeunerenard (approved):

PR #252, opened 2026-04-28
Reuse relay connections across pairings

marcus-pousette-hp:
Previously, each relayed app connection opened its own HyperDHT connection to the same relay. Multiple relayed connections through one relay therefore created duplicate relay transport sockets and extra keepalive/session overhead.

This adds an internal relay pool keyed by relay public key. Relay pairings now reuse one relay transport connection when possible, while each app connection still gets its own blind-relay pairing/raw stream.

Adds coverage for two simultaneous relayed app connections through the same relay. Previously this opened four relay transport sockets: client-to-relay and server-to-relay for each app connection. With reuse, the client and server each keep one relay transport socket, and the two app connections are represented as separate blind-relay pairings over those shared sockets.

### Context 

There are already two related reuse mechanisms, but neither owns blind-relay pairings:

- `ConnectionPool` reuses app-level HyperDHT connections keyed by remote public key. It returns an existing encrypted app stream and handles duplicate app connections.
- `_socketPool` / `reusableSocket` reuses lower-level UDX socket routes.

Relay reuse needs different lifecycle semantics. The shared object is the HyperDHT transport connection to the relay, but each app connection still needs its own blind-relay pairing/token/raw stream. When one app connection upgrades to direct or closes, we must release only that pairing, not destroy the shared relay transport if another pairing is still active.

This PR adds a small internal `RelayPool` for that relay-specific ownership model.

### Unpairing delay

When a relayed app connection upgrades to direct, this PR releases only that relay pairing and keeps the shared relay transport open for any other active pairings.

The unpair is delayed because `remote-changed` is only a local signal. It means this side moved its raw stream to the direct path, but it does not prove the peer has also finished switching direct. Sending `unpair` immediately can tear down the relay leg while the peer still needs it, which showed up as a timeout in the direct-upgrade regression test.

The delay is a pragmatic, non-protocol-changing guard. A longer delay is safer for the upgrade race but keeps relay resources around longer after direct upgrade. A future protocol-level improvement could replace this with an explicit “both sides are direct” release signal before unpairing.

### Race Coverage

This is the tricky part of this PR since we re-using relay connection naturally creates a large flake surface around lifecycle mangement.

This PR includes coverage for the relay-pool cases most likely to regress:

- avoids reusing a relay transport after it has started closing
- keeps active relay pairings alive when another pairing upgrades to direct
- clears/unrefs delayed direct-upgrade unpair timers on pool destroy
- keeps the shared relay transport open when a new pairing starts during another pairing’s delayed unpair window

Disclaimer: PR was co-written with AI.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -101,6 +101,7 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     relayToken: relayThrough ? relay.token() : null,
     relaySocket: null,
     relayClient: null,
+    relayPairing: null,
     relayPaired: false,
     validUpgrade: true,
     relayKeepAlive: opts.relayKeepAlive || 5000
@@ -806,12 +807,16 @@ function relayConnection(c, relayThrough, payload, hs) {
   }
 
   c.relayToken = token
-  c.relaySocket = c.dht.connect(publicKey)
-  c.relaySocket.setKeepAlive(c.relayKeepAlive)
-  c.relayClient = relay.Client.from(c.relaySocket, { id: c.relaySocket.publicKey })
+  c.relayPairing = c.dht._relayPool.pair(publicKey, {
+    isInitiator,
+    token,
+    stream: c.rawStream,
+    keepAlive: c.relayKeepAlive
+  })
+  c.relaySocket = c.relayPairing.socket
   c.relayTimeout = setTimeout(onabort, 15000, null)
 
-  c.relayClient.pair(isInitiator, token, c.rawStream).on('error', onabort).on('data', ondata)
+  c.relayPairing.on('error', onabort).on('data', ondata)
 
   function ondata(remoteId) {
     if (c.relayTimeout) clearRelayTimeout(c)
@@ -822,7 +827,9 @@ function relayConnection(c, relayThrough, payload, hs) {
 
     c.relayPaired = true
 
-    const { remotePort, remoteHost, socket } = c.relaySocket.rawStream
+    // Keep a stable reference since c.relaySocket is cleared after direct upgrade.
+    const relaySocket = c.relaySocket
+    const { remotePort, remoteHost, socket } = relaySocket.rawStream
 
     c.rawStream
       .on('close', () => destroyRelayConnection(c))
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -827,9 +827,7 @@ function relayConnection(c, relayThrough, payload, hs) {
 
     c.relayPaired = true
 
-    // Keep a stable reference since c.relaySocket is cleared after direct upgrade.
-    const relaySocket = c.relaySocket
-    const { remotePort, remoteHost, socket } = relaySocket.rawStream
+    const { remotePort, remoteHost, socket } = c.relaySocket.rawStream
 
     c.rawStream
       .on('close', () => destroyRelayConnection(c))
--- a/lib/relay-connection.js
+++ b/lib/relay-connection.js
@@ -12,9 +12,9 @@ function clearRelayTimeout(connectionOrHandshake) {
 }
 
 function closeRelayConnection(connectionOrHandshake) {
-  const relay = resetRelayConnection(connectionOrHandshake)
+  const pairing = resetRelayConnection(connectionOrHandshake)
 
-  if (relay) closeRelay(relay)
+  if (pairing) pairing.closePairing()
 }
 
 function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging) {
@@ -54,31 +54,20 @@ function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging) 
 }
 
 function destroyRelayConnection(connectionOrHandshake) {
-  const relay = resetRelayConnection(connectionOrHandshake)
+  const pairing = resetRelayConnection(connectionOrHandshake)
 
-  if (relay) destroyRelay(relay)
+  if (pairing) pairing.release()
 }
 
 function resetRelayConnection(connectionOrHandshake) {
   if (connectionOrHandshake.relayTimeout) clearRelayTimeout(connectionOrHandshake)
 
-  const relay = connectionOrHandshake.relayPairing || connectionOrHandshake.relaySocket
+  const pairing = connectionOrHandshake.relayPairing
   // Drop relay references so the app connection no longer keeps relay state alive.
   connectionOrHandshake.relayToken = null
   connectionOrHandshake.relaySocket = null
   connectionOrHandshake.relayClient = null
   connectionOrHandshake.relayPairing = null
 
-  return relay
-}
-
-function closeRelay(relay) {
-  if (relay.closePairing) return relay.closePairing()
-  if (relay.release) return relay.release()
-  relay.end()
-}
-
-function destroyRelay(relay) {
-  if (relay.release) return relay.release()
-  relay.destroy()
+  return pairing
 }
--- a/lib/relay-connection.js
+++ b/lib/relay-connection.js
@@ -12,9 +12,9 @@ function clearRelayTimeout(connectionOrHandshake) {
 }
 
 function closeRelayConnection(connectionOrHandshake) {
-  const socket = resetRelayConnection(connectionOrHandshake)
+  const relay = resetRelayConnection(connectionOrHandshake)
 
-  if (socket) socket.end()
+  if (relay) closeRelay(relay)
 }
 
 function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging) {
@@ -54,19 +54,31 @@ function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging) 
 }
 
 function destroyRelayConnection(connectionOrHandshake) {
-  const socket = resetRelayConnection(connectionOrHandshake)
+  const relay = resetRelayConnection(connectionOrHandshake)
 
-  if (socket) socket.destroy()
+  if (relay) destroyRelay(relay)
 }
 
 function resetRelayConnection(connectionOrHandshake) {
   if (connectionOrHandshake.relayTimeout) clearRelayTimeout(connectionOrHandshake)
 
-  const socket = connectionOrHandshake.relaySocket
+  const relay = connectionOrHandshake.relayPairing || connectionOrHandshake.relaySocket
   // Drop relay references so the app connection no longer keeps relay state alive.
   connectionOrHandshake.relayToken = null
   connectionOrHandshake.relaySocket = null
   connectionOrHandshake.relayClient = null
+  connectionOrHandshake.relayPairing = null
 
-  return socket
+  return relay
+}
+
+function closeRelay(relay) {
+  if (relay.closePairing) return relay.closePairing()
+  if (relay.release) return relay.release()
+  relay.end()
+}
+
+function destroyRelay(relay) {
+  if (relay.release) return relay.release()
+  relay.destroy()
 }
--- a/lib/relay-pool.js
+++ b/lib/relay-pool.js
@@ -52,7 +52,7 @@ class RelayPoolEntry {
     this.socket = pool._dht.connect(publicKey)
     this.client = relay.Client.from(this.socket, { id: this.socket.publicKey })
     this.pairings = new Map()
-    this.pendingReleases = 0
+    this.pendingReleaseTimers = new Set()
     this.destroyed = false
 
     this._onclose = this._onclose.bind(this)
@@ -84,12 +84,14 @@ class RelayPoolEntry {
     if (!this.pairings.delete(pairing.keyString)) return
 
     if (unpair && delay > 0 && !this.destroyed) {
-      this.pendingReleases++
-      setTimeout(() => {
-        this.pendingReleases--
+      const timer = setTimeout(() => {
+        this.pendingReleaseTimers.delete(timer)
         if (!this.destroyed) unpairRelay(this.client, pairing.token)
         this._closeMaybe(unpair)
       }, delay)
+
+      this.pendingReleaseTimers.add(timer)
+      if (timer.unref) timer.unref()
       return
     }
 
@@ -103,6 +105,7 @@ class RelayPoolEntry {
 
     this.destroyed = true
     this.pool._delete(this)
+    this._clearPendingReleaseTimers()
     this.socket.off('close', this._onclose)
     if (graceful) this.socket.end()
     else this.socket.destroy()
@@ -113,6 +116,7 @@ class RelayPoolEntry {
 
     this.destroyed = true
     this.pool._delete(this)
+    this._clearPendingReleaseTimers()
 
     for (const pairing of this.pairings.values()) {
       pairing.detach()
@@ -127,12 +131,18 @@ class RelayPoolEntry {
   }
 
   _closeMaybe(graceful) {
-    if (this.pairings.size === 0 && this.pendingReleases === 0) this.close(graceful)
+    if (this.pairings.size === 0 && this.pendingReleaseTimers.size === 0) this.close(graceful)
+  }
+
+  _clearPendingReleaseTimers() {
+    for (const timer of this.pendingReleaseTimers) clearTimeout(timer)
+    this.pendingReleaseTimers.clear()
   }
 
   _onclose() {
     this.destroyed = true
     this.pool._delete(this)
+    this._clearPendingReleaseTimers()
 
     for (const pairing of this.pairings.values()) {
       pairing.destroy()
--- a/lib/relay-pool.js
+++ b/lib/relay-pool.js
@@ -1,0 +1,203 @@
+const { EventEmitter, once } = require('events')
+const b4a = require('b4a')
+const relay = require('blind-relay')
+
+// Direct upgrade is observed locally first; defer unpairing so the peer can switch too.
+const DIRECT_UPGRADE_UNPAIR_DELAY = 1000
+
+module.exports = class RelayPool {
+  constructor(dht) {
+    this._dht = dht
+    this._entries = new Map()
+  }
+
+  pair(publicKey, opts) {
+    return this._get(publicKey, opts.keepAlive).pair(opts)
+  }
+
+  async destroy() {
+    const closing = []
+
+    for (const entry of this._entries.values()) {
+      closing.push(entry.destroy())
+    }
+
+    await Promise.allSettled(closing)
+  }
+
+  _get(publicKey, keepAlive) {
+    const keyString = b4a.toString(publicKey, 'hex')
+    let entry = this._entries.get(keyString)
+
+    if (!entry || entry.destroyed) {
+      entry = new RelayPoolEntry(this, keyString, publicKey, keepAlive)
+      this._entries.set(keyString, entry)
+    } else {
+      entry.setKeepAlive(keepAlive)
+    }
+
+    return entry
+  }
+
+  _delete(entry) {
+    if (this._entries.get(entry.keyString) === entry) this._entries.delete(entry.keyString)
+  }
+}
+
+class RelayPoolEntry {
+  constructor(pool, keyString, publicKey, keepAlive) {
+    this.pool = pool
+    this.keyString = keyString
+    this.socket = pool._dht.connect(publicKey)
+    this.client = relay.Client.from(this.socket, { id: this.socket.publicKey })
+    this.pairings = new Map()
+    this.pendingReleases = 0
+    this.destroyed = false
+
+    this._onclose = this._onclose.bind(this)
+    this.socket.once('close', this._onclose)
+    this.socket.on('error', noop)
+    this.setKeepAlive(keepAlive)
+  }
+
+  setKeepAlive(keepAlive) {
+    this.socket.setKeepAlive(keepAlive)
+  }
+
+  pair({ isInitiator, token, stream }) {
+    const request = this.client.pair(isInitiator, token, stream)
+    const pairing = new RelayPoolPairing(this, token, request)
+
+    this.pairings.set(pairing.keyString, pairing)
+
+    return pairing
+  }
+
+  release(pairing, unpair, delay = 0) {
+    if (!this.pairings.delete(pairing.keyString)) return
+
+    if (unpair && delay > 0 && !this.destroyed) {
+      this.pendingReleases++
+      setTimeout(() => {
+        this.pendingReleases--
+        if (!this.destroyed) unpairRelay(this.client, pairing.token)
+        this._closeMaybe(unpair)
+      }, delay)
+      return
+    }
+
+    if (unpair && !this.destroyed) unpairRelay(this.client, pairing.token)
+
+    this._closeMaybe(unpair)
+  }
+
+  close(graceful = true) {
+    if (this.destroyed) return
+
+    this.destroyed = true
+    this.pool._delete(this)
+    this.socket.off('close', this._onclose)
+    if (graceful) this.socket.end()
+    else this.socket.destroy()
+  }
+
+  async destroy() {
+    if (this.destroyed) return
+
+    this.destroyed = true
+    this.pool._delete(this)
+
+    for (const pairing of this.pairings.values()) {
+      pairing.detach()
+    }
+
+    this.pairings.clear()
+    this.socket.off('close', this._onclose)
+
+    const closed = this.socket.destroyed ? null : once(this.socket, 'close')
+    this.socket.destroy()
+    if (closed) await closed
+  }
+
+  _closeMaybe(graceful) {
+    if (this.pairings.size === 0 && this.pendingReleases === 0) this.close(graceful)
+  }
+
+  _onclose() {
+    this.destroyed = true
+    this.pool._delete(this)
+
+    for (const pairing of this.pairings.values()) {
+      pairing.destroy()
+    }
+
+    this.pairings.clear()
+  }
+}
+
+class RelayPoolPairing extends EventEmitter {
+  constructor(entry, token, request) {
+    super()
+
+    this.entry = entry
+    this.token = token
+    this.keyString = token.toString('hex')
+    this.request = request
+    this.released = false
+
+    this._ondata = this._ondata.bind(this)
+    this._onerror = this._onerror.bind(this)
+
+    request.on('data', this._ondata)
+    request.on('error', this._onerror)
+  }
+
+  get socket() {
+    return this.entry.socket
+  }
+
+  release() {
+    this._release(true)
+  }
+
+  closePairing() {
+    this._release(true, DIRECT_UPGRADE_UNPAIR_DELAY)
+  }
+
+  detach() {
+    this._release(false)
+  }
+
+  destroy() {
+    this._release(false)
+    this.request.stream.destroy()
+  }
+
+  _release(unpair, delay = 0) {
+    if (this.released) return
+    this.released = true
+
+    this.request.off('data', this._ondata)
+    this.request.off('error', this._onerror)
+    this.request.on('error', noop)
+
+    this.entry.release(this, unpair, delay)
+  }
+
+  _ondata(data) {
+    this.emit('data', data)
+  }
+
+  _onerror(err) {
+    if (this.listenerCount('error') === 0) return
+    this.emit('error', err)
+  }
+}
+
+function noop() {}
+
+function unpairRelay(client, token) {
+  try {
+    client.unpair(token)
+  } catch {}
+}
--- a/lib/relay-pool.js
+++ b/lib/relay-pool.js
@@ -2,8 +2,9 @@ const { EventEmitter, once } = require('events')
 const b4a = require('b4a')
 const relay = require('blind-relay')
 
-// Direct upgrade is observed locally first; defer unpairing so the peer can switch too.
-const DIRECT_UPGRADE_UNPAIR_DELAY = 1000
+// Direct upgrade is observed locally first; keep the relay pairing briefly so
+// the peer can finish switching direct before the relay receives an unpair.
+const DIRECT_UPGRADE_UNPAIR_DELAY = 10000
 
 module.exports = class RelayPool {
   constructor(dht) {
@@ -29,7 +30,7 @@ module.exports = class RelayPool {
     const keyString = b4a.toString(publicKey, 'hex')
     let entry = this._entries.get(keyString)
 
-    if (!entry || entry.destroyed) {
+    if (!entry || !entry.reusable) {
       entry = new RelayPoolEntry(this, keyString, publicKey, keepAlive)
       this._entries.set(keyString, entry)
     } else {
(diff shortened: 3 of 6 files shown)


lib/connect.js
```diff
@@ -814,7 +819,9 @@ function relayConnection(c, relayThrough, payload, hs) {
 
     c.relayPaired = true
 
-    const { remotePort, remoteHost, socket } = c.relaySocket.rawStream
+    // Keep a stable reference since c.relaySocket is cleared after direct upgrade.
+    const relaySocket = c.relaySocket
+    const { remotePort, remoteHost, socket } = relaySocket.rawStream
```
lejeunerenard: Seems to have not gotten the update from #251

lib/relay-pool.js
```diff
@@ -0,0 +1,183 @@
...
+  _delete(entry) {
+    if (this._entries.get(entry.keyString) === entry) this._entries.delete(entry.keyString)
+  }
+}
+
+class RelayPoolEntry {
+  constructor(pool, keyString, publicKey, keepAlive) {
+    this.pool = pool
+    this.keyString = keyString
+    this.socket = pool._dht.connect(publicKey)
```
lejeunerenard, line 59: We aren't using the connection pool atm if there's one available. This is fine as the pool isn't used before either. If we did want to support pooling the connection with `ConnectionPool`, is that possible?

lib/connect.js
```diff
@@ -798,12 +799,16 @@ function relayConnection(c, relayThrough, payload, hs) {
...
 
   c.relayToken = token
-  c.relaySocket = c.dht.connect(publicKey)
-  c.relaySocket.setKeepAlive(c.relayKeepAlive)
-  c.relayClient = relay.Client.from(c.relaySocket, { id: c.relaySocket.publicKey })
+  c.relayPairing = c.dht._relayPool.pair(publicKey, {
+    isInitiator,
+    token,
+    stream: c.rawStream,
+    keepAlive: c.relayKeepAlive
```
lejeunerenard, line 814: How is this respected when the connection is reused? Currently it looks like new pairings clobber the old connection's keepAlive value.

lib/relay-connection.js
```diff
@@ -8,25 +8,37 @@ function clearRelayTimeout(connectionOrHandshake) {
...
 
-  if (socket) socket.destroy()
+  if (relay) destroyRelay(relay)
 }
 
 function resetRelayConnection(connectionOrHandshake) {
   if (connectionOrHandshake.relayTimeout) clearRelayTimeout(connectionOrHandshake)
 
-  const socket = connectionOrHandshake.relaySocket
+  const relay = connectionOrHandshake.relayPairing || connectionOrHandshake.relaySocket
```
lejeunerenard: If the relaySocket is returned, that will throw when calling `closeRelay()` or `destroyRelay()` right? Not sure it's possible to have a relaySocket but not a relayPairing anyways.

lib/relay-connection.js
```diff
@@ -8,25 +8,37 @@ function clearRelayTimeout(connectionOrHandshake) {
...
   connectionOrHandshake.relayClient = null
+  connectionOrHandshake.relayPairing = null
 
-  return socket
+  return relay
+}
+
+function closeRelay(relay) {
+  if (relay.closePairing) return relay.closePairing()
+  if (relay.release) return relay.release()
```
lejeunerenard: In what situation will a relay have a `release()` but not a `closePairing()`?

lib/server.js
```diff
@@ -652,37 +653,40 @@ module.exports = class Server extends EventEmitter {
...
 
-        if (hs.prepunching) clearTimeout(hs.prepunching)
-        hs.prepunching = null
+      if (hs.prepunching) clearTimeout(hs.prepunching)
+      hs.prepunching = null
 
-        const { remotePort, remoteHost, socket } = hs.relaySocket.rawStream
+      // Keep a stable reference since hs.relaySocket is cleared after direct upgrade.
+      const relaySocket = hs.relaySocket
+      const { remotePort, remoteHost, socket } = relaySocket.rawStream
```
lejeunerenard: This also didn't get the update.

lejeunerenard (changes requested):

lib/relay-connection.js
marcus-pousette-hp: Yeah, that fallback was leftover from testing/refactoring while I was moving the cleanup path from socket-owned to pairing-owned. In the current pooled path relayPairing is always created together with the relaySocket, and the pairing is the actual cleanup owner. I removed the redundant fallback and made the helper call the pairing methods directly.

lib/relay-connection.js
marcus-pousette-hp: Se my other comment https://github.com/holepunchto/hyperdht/pull/252#discussion_r3287631397 and fix https://github.com/holepunchto/hyperdht/pull/252/commits/1336aa7c24286b1f4ec67d802c736db12885d070

lib/connect.js
marcus-pousette-hp: Good catch. Fixed in https://github.com/holepunchto/hyperdht/pull/252/commits/cda111194d47e1bb0e4b085727d97b26a081835a I know key bey keepAlive too. An option would be to use keepAlive as the smallest value, but that would lead to behavioural changes and unintended effects if the callers want be smart about the keepAlive option

marcus-pousette-hp: One related thing I noticed while looking at this: relayKeepAlive currently uses opts.relayKeepAlive || 5000 on both the connect and server paths, so relayKeepAlive: false and relayKeepAlive: 0 both end up as 5000. I left that unchanged here since it was pre-existing behavior and this PR is only meant to change relay transport reuse, but it does feel a bit different from connectionKeepAlive, which explicitly supports false. 

Do you think relayKeepAlive should keep the current defaulting behavior, or should we open a follow-up PR (or modify this one) to make false/0 disable relay keepalive like connectionKeepAlive does? @lejeunerenard

lib/relay-pool.js
marcus-pousette-hp: So the two pools reduce duplication at different levels:

ConnectionPool: duplicate app streams to same peer
RelayPool: duplicate relay transport sockets to same relay

An example is 

App opens connections to different peers through the same relay. ConnectionPool would not help because the app targets are different. RelayPool can still reuse the same relay transport because the shared thing is the relay node, not the target peer.

we could in theory have a generalized pool, but it would need to handle both app-stream dedupe and relay transport leases. And it would be easy to blur relay-specific lifecycle handling if we would try to generalize, and in practice I think the solution would be more complicated to maintain and do correctly

lib/connect.js
marcus-pousette-hp: Not sure if I fully understand what you are asking for here, but if is it about the fallback behaviour, it is now removed  https://github.com/holepunchto/hyperdht/pull/252#discussion_r3287631397 and fix https://github.com/holepunchto/hyperdht/commit/1336aa7c24286b1f4ec67d802c736db12885d070

lib/server.js
marcus-pousette-hp: Not sure if I fully understand what you are asking for here, but if is it about the fallback behaviour, it is now removed  https://github.com/holepunchto/hyperdht/pull/252#discussion_r3287631397 and fix https://github.com/holepunchto/hyperdht/commit/1336aa7c24286b1f4ec67d802c736db12885d070

lib/connect.js
lejeunerenard: Sorry should have been more clear, I'm referring to the fact that this line changed in #251 which this PR was originally based on. But in #251 it was fixed as we realized we didn't want to keep the reference around and accounted for it still. This PR didn't get that change.

tl;dr we don't need this comment nor the `relaySocket` variable as the reference doesnt need to be kept in this scope.

lib/relay-pool.js
lejeunerenard: Keeping them separate currently means that if an app uses the same `blind-relay` public key for whatever app level logic, a new and separate stream will be created to the same `blind-relay` because on this line. We do not support currently passing a pool to the `dht.connect()` here, which is fine for now. Just thinking in the future it'd be better to dedup the stream to the blind relay so there is only one.

lib/connect.js
marcus-pousette-hp: aaaah! yeah makes sense now. I missed this thanks !

lib/relay-pool.js
marcus-pousette-hp: Makes sense. So the remaining duplication case is if user/app code also has an app-level stream to the same blind-relay public key; this PR only dedupes relay transports created by RelayPool, not those external app-level streams. I agree that sharing that underlying stream would require plumbing a pool model into this internal dht.connect() path so the relay transport and app-level use can safely reason about ownership. But this could perhaps be a nice improvement PR after this PR?

lejeunerenard: Yeah definitely not now. We can add later if it's ever needed.

lib/connect.js
lejeunerenard: I think we should keep the default behavior as is. Setting the keep alive to `0` would send a message every tick which seems undesirable.

I'm not 100% sure that we should be keying base on keep alive. I'm struggling to think of scenarios where the keep alive would be set to something bigger for a good reason. If we take the smaller the connection is more likely to be kept alive. Using for something else feels more like an abuse of the API than a valid usecase. Was there as scenario you were thinking of?

test/relaying.js
```diff
@@ -521,6 +591,544 @@ test('relay connections through node, client and server side', async function (t
...
+      relayThrough: relayTransportServer.publicKey
+    })
+  ]
+
+  await Promise.all([
+    relaySocketsOpened,
+    relayStreamsPaired,
+    serverSocketsOpened,
+    ...clientSockets.map((socket) => once(socket, 'open'))
+  ])
```
```suggestion
  await Promise.all(clientSockets.map((socket) => once(socket, 'open'))
```
These aren't necessary as since the client connection can't open until the connection is fully established (handshake etc). Timing only matters if we use the server sockets array immediately since it might not have run its callback yet. But we dont until after replies etc.

This line can also will go away if we opt to await the connections individually.

lib/connect.js
marcus-pousette-hp: > I think we should keep the default behavior as is. Setting the keep alive to 0 would send a message every tick which seems undesirable.

Yes! Or 0 could also mean it is disabled somehow in practice. But then I will keep the curent behaviour then!

> I'm not 100% sure that we should be keying base on keep alive. I'm struggling to think of scenarios where the keep alive would be set to something bigger for a good reason. If we take the smaller the connection is more likely to be kept alive. Using for something else feels more like an abuse of the API than a valid usecase. Was there as scenario you were thinking of?

I went down this path a bit when doing the PR there is a tradeoff. The current solution this PR offers preserves the current behaviour, whereas the solution where we would pick the smaller one, would in practice mean callers initial intent of having it big would be ignored.  I don’t have a strong concrete case where a larger relay keepalive must be preserved independently excapt that reducing keepalive traffic/wakeups for callers that intentionally picked a longer interval.

So the path forward here is either perserve the existing behaivour bying keying on keepAlive OR make a simpler solution where use the smaller value, and accept a behavioural change. 

(Having a path where we always choose the larger value is not a safe choice, since another caller might have a real reason for picking a small value). 

Potentially, a third option would by some kind of hybrid where it does not key by default but if the user wants it can "key-it" to not make it affect other settings

lib/relay-pool.js
```diff
@@ -0,0 +1,214 @@
...
```
noahlevenson, line 59: Style thoughts: I don't love a constructor with network I/O side effects. Not everyone shares my vision for the world, though. I might at least reconsider the name of the class. "Entry" sounds like passive state, but this doohickey starts dialing people on the internet when you instantiate it. It's probably just a `RelayPoolConnection`, isn't it?

lib/relay-pool.js
```diff
@@ -0,0 +1,214 @@
...
+    this.pendingReleaseTimers = new Set()
+    this.destroyed = false
+
+    this._onclose = this._onclose.bind(this)
+    this.socket.once('close', this._onclose)
+    this.socket.on('error', noop)
+    this.socket.setKeepAlive(keepAlive)
+  }
+
+  get reusable() {
```
noahlevenson, line 72: Reviewing at the request of @marcus-pousette-hp...

The death signal is actually 4 async signals spread across different layers of abstraction. Scares me a little. The different destruction paths and layers of destruction state (`RelayPool` vs. `socket` vs. `client`) make it hard to reason about whether there might be a subtle race between the predicates. Feels like a violation of the isolation of complexity.

lib/relay-pool.js
marcus-pousette-hp: Good point. I agree with you. https://github.com/holepunchto/hyperdht/pull/252/commits/5d6fed89e32dde445ab6925dafbff3bb448cfbc6 
I did the rename you suggested, makes more sense here

lib/relay-pool.js
marcus-pousette-hp: Makes sense, yeap this could be fragile even though it is tested. I did a cleanup commit

https://github.com/holepunchto/hyperdht/pull/252/commits/13af091ecd544a3cc212daed438a40e62556d952 

that unifies this somewhat, and makes this a bit easier to overview

lib/connect.js
lejeunerenard: I think that we shouldn't key by keep alive and take the min of existing keep alive values. Even if we keyed by keep alive the scenario where someone wants to keep the keep alive longer to reduce traffic would mean more traffic in total as there are now two connections to the blind relay and each would have their own keep alive messages. I'd rather only keep one connection and take the safer keep alive value.

lib/relay-pool.js
lejeunerenard: I agree with the idea of renaming. Much clearer that there's a side effect. Related there are still variables with the name `entry`/`entries`, should these be "connection"s instead?

test/relaying.js
```diff
@@ -212,6 +214,74 @@ async function waitFor(fn, timeout = 2000) {
   }
 }
 
+function getOnlyRelayPoolConnection(node) {
```
lejeunerenard: All helpers should be at the end of the file unless within a given test.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+  const clientSockets = [
+    clientNode.connect(appServer.publicKey, {
+      localConnection: false,
+      relayThrough: relayTransportServer.publicKey
+    }),
+    clientNode.connect(appServer.publicKey, {
+      localConnection: false,
+      relayThrough: relayTransportServer.publicKey
+    })
+  ]
```
lejeunerenard: These should probably just be separated instead of defined as an array. They are used individually anyways and avoids using `Promise.race()` like so:
```js
const firstServerClosed = Promise.race(serverSockets.map((socket) => once(socket, 'close')))
```
If we await them individually, we know the orders of the server sockets too and the logic in the test is a bit clearer as each connection (either client or socket) is labeled.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+  ])
+
+  t.is(relaySockets.length, 2, 'both peers reuse one transport socket to the relay')
+
+  const replies = clientSockets.map((socket) => once(socket, 'data'))
+  clientSockets[0].write(Buffer.from('hello 1'))
+  clientSockets[1].write(Buffer.from('hello 2'))
+
+  t.alike((await replies[0])[0], Buffer.from('hello 1'), 'first relayed connection carries data')
+  t.alike((await replies[1])[0], Buffer.from('hello 2'), 'second relayed connection carries data')
```
```suggestion
  t.alike(await replies[1], [Buffer.from('hello 2')], 'second relayed connection carries data')
```
Especially if the replies are separated per [this comment](https://github.com/holepunchto/hyperdht/pull/252/changes#r3351167993), the assertion becomes more direct without two array indexes etc.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+
+  const firstServerClosed = Promise.race(serverSockets.map((socket) => once(socket, 'close')))
+  await endAndCloseSocket(clientSockets[0])
+  await firstServerClosed
+  await firstPairingReleased
```
lejeunerenard: Can't we only wait on `firstPairingReleased`? A pairing should only release after two relay streams are closed which should be the client & server as orchestrated by this test?

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+  t.is(closedRelaySockets, 2, 'shared relay transports close after all pairings close')
+
+  const reconnectedServerSocket = waitFor(() => serverSockets.length === 3)
+  const reconnectedClientSocket = clientNode.connect(appServer.publicKey, {
+    localConnection: false,
+    relayThrough: relayTransportServer.publicKey
+  })
+
+  await Promise.all([once(reconnectedClientSocket, 'open'), reconnectedServerSocket])
+  await waitFor(() => relaySockets.length === 4)
```
lejeunerenard: Same as noted [here](https://github.com/holepunchto/hyperdht/pull/252/changes#r3307557528) we only need to await the `open` event since we're forcing relayed connections.

The `waitFor()` is also unnecessary as we test the number of sockets on the next line.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+      localConnection: false,
+      relayThrough: relayTransportServer.publicKey
+    })
+  ]
+
+  await Promise.all([
+    relaySocketsOpened,
+    serverSocketsOpened,
+    ...clientSockets.map((socket) => once(socket, 'open'))
+  ])
```
```suggestion
  await Promise.all(clientSockets.map((socket) => once(socket, 'open')))
```
Same as [here](https://github.com/holepunchto/hyperdht/pull/252/changes#r3307557528) (minus the point about the connection array).

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+  t.is(serverNode._relayPool._entries.size, 0, 'server relay pool entry is removed')
+
+  const reconnectedServerSocket = waitFor(() => serverSockets.length === 3)
+  const reconnectedClientSocket = clientNode.connect(appServer.publicKey, {
+    localConnection: false,
+    relayThrough: relayTransportServer.publicKey
+  })
+
+  await Promise.all([once(reconnectedClientSocket, 'open'), reconnectedServerSocket])
+  await waitFor(() => relaySockets.length === 4)
```
lejeunerenard: Same as [here](https://github.com/holepunchto/hyperdht/pull/252/changes#r3351234009).

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+  })
+
+  await waitFor(() => relay._pairing.size === 1)
+  t.is(relay._pairing.size, 1, 'relay has one pending half-pairing')
+
+  pairing.release()
+  rawStream.destroy()
+
+  await waitFor(() => relay._pairing.size === 0)
+  await waitFor(() => clientNode._relayPool._entries.size === 0)
```
```suggestion
```

This can be removed since the client's entries should be empty immediately since its the one releasing.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+  t.is(relay._pairing.size, 0, 'pending half-pairing is unpaired')
```
```suggestion
  t.pass('pending half-pairing is unpaired')
```
Can be a `.pass()` since the assertion check is in the `waitFor()` above.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+
+  const thirdPairing = pair(1000)
+
+  t.is(thirdPairing.socket, firstSocket, 'same keepalive value reuses transport')
+  t.is(clientNode._relayPool._entries.size, 2, 'same keepalive value reuses existing entry')
+
+  for (const { pairing, stream } of pairings) {
+    pairing.release()
+    stream.destroy()
+  }
```
lejeunerenard: Releasing the pairing and destroying the streams isn't necessary, right? They all get released and destroyed when the node is destroyed.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+  const delayedUnpair = captureDelayedRelayUnpairTimer(t)
+
+  // closePairing is the direct-upgrade cleanup path that schedules delayed unpair.
+  pairing.closePairing()
+  rawStream.destroy()
+
+  t.ok(delayedUnpair.unrefed, 'delayed unpair timer is unrefed')
+
+  await clientNode._relayPool.destroy()
+  t.ok(delayedUnpair.cleared, 'delayed unpair timer is cleared on pool destroy')
```
lejeunerenard: Monkey patching `setTimeout()` & `clearTimeout()` isn't required for this test. You can check the presence of the timer and that its unrefed like so:
```js
  t.is(pairing.entry.pendingReleaseTimers.size, 1, 'added a timer')
  t.absent(timer.hasRef(), 'delayed unpair timer is unrefed')
```

To check clearing on the pool being destroyed, check that the `pendingReleaseTimers` is empty since that's cleared along withe the timers.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+    keepAlive: 5000
+  })
+
+  await waitFor(() => relay._pairing.size === 1)
+
+  const relaySocket = firstPairing.socket
+  const delayedUnpair = captureDelayedRelayUnpairTimer(t)
+
+  // Simulate direct upgrade for the first pairing, leaving its unpair delayed.
+  firstPairing.closePairing()
```
lejeunerenard: This test also doesn't need monkey patching. The monkey patching in this case is to reduce the time for the timer and to control when it fires. These can be done by:
1. Use `firstPairing._release(true, 1000)` instead of `closePairing()` to simulate a shorter version. The custom delay can be pretty short since everything is synchronous afterwards except waiting for the pairing to register which should be short and well under something like 1s.
2. You can ensure the timer fired before proceeding by using the second `waitFor()` condition since it checks that the timer fired and removed the pairing.

test/relaying.js
```diff
@@ -521,6 +591,548 @@ test('relay connections through node, client and server side', async function (t
...
+    relaySocket,
+    'same relay transport remains pooled'
+  )
+  t.absent(relaySocket.destroying, 'relay transport is not closing')
+
+  secondPairing.release()
+  secondStream.destroy()
+
+  await waitFor(() => relay._pairing.size === 0)
+  await waitFor(() => clientNode._relayPool._entries.size === 0)
```
lejeunerenard: Why is this necessary? There are no asserts afterwards and this is done via destroying the nodes.

test/relaying.js
```diff
@@ -633,6 +1245,189 @@ test('relay connection upgrades to direct connection', async function (t) {
...
+  upgradeClientSocket.on('error', (err) => t.comment(err.message))
+  relayOnlyClientSocket.on('error', (err) => t.comment(err.message))
+
+  const [upgradeServerSocket, relayOnlyServerSocket] = await Promise.all([
+    upgradeServerSocketOpened,
+    relayOnlyServerSocketOpened,
+    once(upgradeClientSocket, 'open'),
+    once(relayOnlyClientSocket, 'open')
+  ])
+  await Promise.all([relaySocketsOpened, relayStreamsPaired])
```
```suggestion
  await Promise.all(
    once(upgradeClientSocket, 'open'),
    once(relayOnlyClientSocket, 'open')
  ])
```
Since `upgradeServerSocket` & `relayOnlyServerSocket` are the only sockets on their respective servers and aren't needed immediately, these can be set in their respective callbacks. The unused promises etc can be cleaned up then.

test/relaying.js
```diff
@@ -212,6 +214,74 @@ async function waitFor(fn, timeout = 2000) {
   }
 }
 
+function getOnlyRelayPoolConnection(node) {
+  const entries = [...node._relayPool._entries.values()]
+  if (entries.length !== 1) throw new Error('Expected exactly one relay pool connection')
+  return entries[0]
+}
+
+function captureDelayedRelayUnpairTimer(t) {
```
lejeunerenard: Can be removed if the timers are tested as suggested below.

lejeunerenard (changes requested):

test/relaying.js
marcus-pousette-hp: I agree, that makes it nicer. 

Fixed in https://github.com/holepunchto/hyperdht/commit/3f91591a839be84f9edac1f4d0ef6504ed1a0b79

test/relaying.js
marcus-pousette-hp: Fixed part of https://github.com/holepunchto/hyperdht/commit/3f91591a839be84f9edac1f4d0ef6504ed1a0b79

lib/connect.js
marcus-pousette-hp: I agree with you. Actually, given that this PR will already give some saving justifies this change. 

Fixed in https://github.com/holepunchto/hyperdht/commit/e2ccc394a52d876234d95f0097ab9736b7fb2f25

lib/relay-pool.js
marcus-pousette-hp: Good catch, I misesd this. Fixed https://github.com/holepunchto/hyperdht/commit/0f7cbda02462076e78c63bbd150b1deef3ccd2ba

test/relaying.js
marcus-pousette-hp: Renaming fix in https://github.com/holepunchto/hyperdht/commit/0f7cbda02462076e78c63bbd150b1deef3ccd2ba

test/relaying.js
marcus-pousette-hp: Fixed in https://github.com/holepunchto/hyperdht/commit/3f91591a839be84f9edac1f4d0ef6504ed1a0b79

test/relaying.js
marcus-pousette-hp: Fixed in https://github.com/holepunchto/hyperdht/commit/0ed08f03cf216caedf308daf99ea849e0f895770

I agree. The monkey patch was remnants from earlier behaviour, which could be made way easier now

test/relaying.js
marcus-pousette-hp: Yea great point!! 

Fixed in https://github.com/holepunchto/hyperdht/commit/a8fa8b1f998c55480e4d8e4044dc3e544ad88586

This now uses `_release(true, 1000)` to simulate the delayed direct-upgrade unpair, and waits for the relay pairing count to drop from 2 to 1 to prove the timer fired while the second pairing stayed alive.

test/relaying.js
marcus-pousette-hp: In the earlier version those waits were implicit cleanup assertions after releasing the second pairing. That was outside this test’s main scope, so I removed that part and kept the test focused on the delayed-unpair race.

test/relaying.js
marcus-pousette-hp: Yup good point, fixed
https://github.com/holepunchto/hyperdht/commit/88b548747c6d70e21f33b41ff5c819497cc4b701

test/relaying.js
marcus-pousette-hp: Fixed

test/relaying.js
lejeunerenard: The `relaySocketsOpened` & `relayStreamsPaired` are still not required here:
https://github.com/holepunchto/hyperdht/pull/252/changes#diff-77e8658bbd4f0803c631d07808c79139fb0bb8f533f28dd5268a9bc196613b13R622

We know they are true because the client socket's being `open` when its forced to relay means the sockets were open and the streams have paired.

test/relaying.js
lejeunerenard: So these aren't required still correct? I only see the `relay._pairing.size === 0` removed at the moment.

test/relaying.js
lejeunerenard: Thanks the sockets are now defined in their callbacks, but another aspect of this suggested change was removing the `relaySocketsOpened` and the `relayStreamsPaired` as those are not required because the client socket's being `open` implies they were done per this comment:
https://github.com/holepunchto/hyperdht/pull/252#discussion_r3358162901

test/relaying.js
lejeunerenard: Looks like while there are less helpers, there are still helpers added in the middle of the tests.

test/relaying.js
marcus-pousette-hp: Done in 7320914. firstPairingReleased already waits for both relay streams in that pairing to close, so the extra server socket close wait was redundant.

test/relaying.js
marcus-pousette-hp: Done in 5985847

test/relaying.js
marcus-pousette-hp: Done in 73ee732

test/relaying.js
marcus-pousette-hp: Also handled in 5985847

test/relaying.js
marcus-pousette-hp: Done in adeb526

test/relaying.js
marcus-pousette-hp: I agree, this makes sense. Done in adeb526

test/relaying.js
marcus-pousette-hp: Yup this is redundnat, the test was overly defensive. Done in 2a65785

test/relaying.js
marcus-pousette-hp: Yeah you are right here too. Fixed 5b4b5cf

test/relaying.js
marcus-pousette-hp: Missed that. Now fixed in 0ce3a5a.

test/relaying.js
marcus-pousette-hp: Yeah, hmm for some reason I only removed one part by mistake. Fixed in bafe3fd

test/relaying.js
marcus-pousette-hp: Fixed in 4e00f13.

marcus-pousette-hp: Makes things way simpler, thank you

lib/relay-pool.js
```diff
@@ -0,0 +1,226 @@
...
+      connection.setKeepAlive(keepAlive)
+    }
+
+    return connection
+  }
+
+  _delete(connection) {
+    if (this._connections.get(connection.keyString) === connection) {
+      this._connections.delete(connection.keyString)
+    }
```
lejeunerenard, line 51: Just wanted to note: Technically a connection could be orphaned here if its not reusable when another connection for the same public key is made.

marcus-pousette-hp: You are right, good catch! Fixed https://github.com/holepunchto/hyperdht/pull/252/commits/e4c290802d6254a84f8c08b4da352956bc7fc5ab

lib/relay-pool.js
```diff
@@ -0,0 +1,231 @@
...
+
+    if (graceful) this.socket.end()
+    else this.socket.destroy()
+  }
+
+  async destroy() {
+    if (!this._closeState()) return
+
+    this._releasePairings(false)
+    const closed = this.socket.destroyed ? null : once(this.socket, 'close')
```
mafintosh: Events.once does weird stuff if error event is emitted. Just listen once for close only manually

marcus-pousette-hp: Good point! that was subtle and fragile.. Fixed https://github.com/holepunchto/hyperdht/pull/252/commits/18376ef6ab85b04abf3f92214fe12c08c6ced4bf

PR #254, opened 2026-05-05
Bump `compact-encoding` to `3.0.0`

lejeunerenard:
For the encoding using `compact-encoding` all `c.buffer` were converted to `c.optionalBuffer`.

Depends on:

- https://github.com/holepunchto/hypercore-crypto/pull/25
- https://github.com/holepunchto/hyperdht-address/pull/2
- https://github.com/holepunchto/dht-rpc/pull/127
- https://github.com/mafintosh/bogon/pull/5
- https://github.com/holepunchto/blind-relay/pull/16

### DIFF
--- a/lib/messages.js
+++ b/lib/messages.js
@@ -425,20 +425,31 @@ exports.pluginRequest = {
     c.string.preencode(state, m.plugin)
     c.uint.preencode(state, m.version)
     c.uint.preencode(state, m.command)
-    c.buffer.preencode(state, m.value)
+    state.end++ // max flag is 1 so always one byte
+
+    if (m.value) c.buffer.preencode(state, m.value)
   },
   encode(state, m) {
+    const flags = m.value ? 1 : 0
+
     c.string.encode(state, m.plugin)
     c.uint.encode(state, m.version)
     c.uint.encode(state, m.command)
-    c.buffer.encode(state, m.value)
+    c.uint.encode(state, flags)
+
+    if (m.value) c.buffer.encode(state, m.value)
   },
   decode(state) {
+    const r0 = c.string.decode(state)
+    const r1 = c.uint.decode(state)
+    const r2 = c.uint.decode(state)
+    const flags = c.uint.decode(state)
+
     return {
-      plugin: c.string.decode(state),
-      version: c.uint.decode(state),
-      command: c.uint.decode(state),
-      value: c.buffer.decode(state)
+      plugin: r0,
+      version: r1,
+      command: r2,
+      value: (flags & 1) !== 0 ? c.buffer.decode(state) : null
     }
   }
 }
--- a/lib/messages.js
+++ b/lib/messages.js
@@ -425,20 +425,20 @@ exports.pluginRequest = {
     c.string.preencode(state, m.plugin)
     c.uint.preencode(state, m.version)
     c.uint.preencode(state, m.command)
-    c.optionalBuffer.preencode(state, m.value)
+    c.buffer.preencode(state, m.value)
   },
   encode(state, m) {
     c.string.encode(state, m.plugin)
     c.uint.encode(state, m.version)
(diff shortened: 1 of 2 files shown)


lib/messages.js
```diff
@@ -403,18 +403,18 @@ exports.mutablePutRequest = {
 exports.mutableGetResponse = {
   preencode(state, m) {
     c.uint.preencode(state, m.seq)
-    c.buffer.preencode(state, m.value)
+    c.optionalBuffer.preencode(state, m.value)
```
mafintosh: required buffer

lib/messages.js
```diff
@@ -381,20 +381,20 @@ exports.mutablePutRequest = {
...
-    c.buffer.encode(state, m.value)
+    c.optionalBuffer.encode(state, m.value)
     c.fixed64.encode(state, m.signature)
   },
   decode(state) {
     return {
       publicKey: c.fixed32.decode(state),
       seq: c.uint.decode(state),
-      value: c.buffer.decode(state),
+      value: c.optionalBuffer.decode(state),
```
mafintosh: required buffer

lib/messages.js
```diff
@@ -363,16 +363,16 @@ exports.announce = {
...
   encode(state, m) {
     c.uint.encode(state, m.seq)
-    c.buffer.encode(state, m.value)
+    c.optionalBuffer.encode(state, m.value)
   },
   decode(state) {
     return {
       seq: c.uint.decode(state),
-      value: c.buffer.decode(state)
+      value: c.optionalBuffer.decode(state)
```
mafintosh: required buffer

lib/messages.js
lejeunerenard: Looks like elsewhere its assumed to be potentially `null`
https://github.com/holepunchto/hyperdht/blob/74dea50245d96dea299aac6fee55f5a01ebc092e/lib/persistent.js#L194

lib/messages.js
lejeunerenard: ~Is checked to be truthy here so seems also potentially `null`:
https://github.com/holepunchto/hyperdht/blob/74dea50245d96dea299aac6fee55f5a01ebc092e/lib/persistent.js#L179~

lejeunerenard: Nevermind, that's not a check on the `value` but decoding as a whole.

lib/messages.js
lejeunerenard: While this check implies that `value` can be `null`, it can't because when the local mutable record is set, it needs to pass this check:
https://github.com/holepunchto/hyperdht/blob/74dea50245d96dea299aac6fee55f5a01ebc092e/lib/persistent.js#L187
so is always required.

lib/messages.js
```diff
@@ -29,14 +29,14 @@ const ipv6Array = c.array(ipv6)
...
-    c.buffer.preencode(state, m.noise)
+    c.optionalBuffer.preencode(state, m.noise)
   },
   encode(state, m) {
     const flags = (m.peerAddress ? 1 : 0) | (m.relayAddress ? 2 : 0)
 
     c.uint.encode(state, flags)
     c.uint.encode(state, m.mode)
-    c.buffer.encode(state, m.noise)
+    c.optionalBuffer.encode(state, m.noise)
```
mafintosh: required

lib/messages.js
```diff
@@ -228,23 +228,23 @@ exports.holepunch = {
...
+    c.optionalBuffer.encode(state, m.payload)
     if (m.peerAddress) ipv4.encode(state, m.peerAddress)
   },
   decode(state) {
     const flags = c.uint.decode(state)
     return {
       mode: c.uint.decode(state),
       id: c.uint.decode(state),
-      payload: c.buffer.decode(state),
+      payload: c.optionalBuffer.decode(state),
```
mafintosh: required

Issue #112, opened 2022-12-08
Unable to holepunch from the local network to a Docker container


louneskmt:
Hey!

I am not able to connect to a DHT server running in a Docker container, from the same local network. To be clear, the Docker stack has its own Docker network, which lives in my home network. Here is a summary of the situation:

![telegram-cloud-photo-size-4-5805510472735242539-y](https://user-images.githubusercontent.com/38882571/206452726-a0b235a2-9f84-4ec7-a458-2c835efd2edf.jpg)

Apparently, it may be because as both the clients and the server have the same external IP `X.X.X.X`, it tries to connect via LAN, but in reality the server is in a container with its own network.

<details>
<summary>Here is the full error, client-side</summary>

```console
Waiting for node to be fully bootstrapped to collect info...

Node info:
- remote host: X.X.X.X
- remote port: 56888
- firewalled: true
- nat type: consistent

Connecting to test server...
Sending 32 bytes for the server to echo
/home/louneskmt/test/node_modules/.pnpm/@hyperswarm+dht@6.3.3/node_modules/@hyperswarm/dht/lib/connect.js:154
  const onabort = () => c.encryptedSocket.destroy(HOLEPUNCH_ABORTED())
                                                  ^

DHTError: HOLEPUNCH_ABORTED: Holepunch aborted
    at Holepuncher.onabort (/home/louneskmt/test/node_modules/.pnpm/@hyperswarm+dht@6.3.3/node_modules/@hyperswarm/dht/lib/connect.js:154:51)
    at Holepuncher.destroy (/home/louneskmt/test/node_modules/.pnpm/@hyperswarm+dht@6.3.3/node_modules/@hyperswarm/dht/lib/holepuncher.js:279:31)
    at Holepuncher._autoDestroy (/home/louneskmt/test/node_modules/.pnpm/@hyperswarm+dht@6.3.3/node_modules/@hyperswarm/dht/lib/holepuncher.js:267:31)
    at Holepuncher._consistentProbe (/home/louneskmt/test/node_modules/.pnpm/@hyperswarm+dht@6.3.3/node_modules/@hyperswarm/dht/lib/holepuncher.js:218:10)
Emitted 'error' event on NoiseSecretStream instance at:
    at WritableState.afterDestroy (/home/louneskmt/test/node_modules/.pnpm/streamx@2.12.5/node_modules/streamx/index.js:444:19)
[...]

PR #259, opened 2026-05-21
Keep LAN shortcut from blocking holepunch fallback

### LINKED ISSUE #210: HyperDHT can not punch through firewall when both server and client are on the same network.
**Issue** : HyperDHT can not punch through firewall when both the client and server are on the same network and the host is a Linux machine using [ufw/gufw](https://itsfoss.com/set-up-firewall-gufw/) firewall.

I have tested this with Linux Mint but this also occurs on Ubuntu and other Linux based OSes.

Reproduction Steps:

1. Start a HyperDHT server on Linux Mint or Ubuntu
2. Make sure in firewall settings, Incoming is set to "Deny"
3. Now try starting a HyperDHT client and it will not punch through, but if you set the button to "Allow" the connection will go through.

marcus-pousette-hp:
When two peers appear to be on the same network, `connect()` currently tries the server's LAN address first. If that LAN ping fails, the connection aborts with `HOLEPUNCH_ABORTED` before the normal holepunch path gets a chance to run.

That means LAN cases with firewalls/default-deny incoming rules can fail even when the regular UDX holepunch path through the observed NAT address would work.

This changes the LAN shortcut to be opportunistic when the remote is holepunchable:

- still try the LAN ping as a fast path
- do not let a failed LAN ping abort the connection before normal holepunch
- keep the previous abort behavior when the remote is not holepunchable and there is no fallback path
- keep the server-side holepunch payload/puncher setup available for same-LAN peers

## Testing

I added testing for the fallback behaivour in the PR. 

But also did a LAN/hairpin repro:

I also tested the same failure shape as #210 with a Docker-based LAN/NAT setup:

- two peers on the same simulated LAN
- both peers behind the same simulated NAT
- direct LAN UDP to the server blocked
- `ufw default deny incoming` enabled on the peers
- hairpin NAT available through the externally observed address

On `origin/main`, the client failed with `HOLEPUNCH_ABORTED` before the normal holepunch hook ran.

On this branch, the failed LAN ping no longer stopped the connection. The client reached the normal holepunch path and successfully exchanged data with the server.

I also ran this repro inside an Ubuntu 24.04 Lima VM with rootful Docker, so the UFW/firewall behavior was tested in a Linux environment.

If needed I can provide the repro scripts used for this. 

------ 

Ref and praise to  https://github.com/holepunchto/hyperdht/pull/236 for the initial solution attempt. 

Addresses the original failure mode reported in #112: when peers appear same-LAN, a failed LAN shortcut can abort with HOLEPUNCH_ABORTED before normal holepunch gets a chance to run. Closes https://github.com/holepunchto/hyperdht/issues/210

------ 

Co-Written with AI

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -257,6 +257,8 @@ async function holepunch(c, opts) {
       if (!remoteHolepunchable) return
 
       function onlanconnect() {
+        // LAN and holepunch completions both converge through onsocket,
+        // which claims c.rawStream once and ignores later winners.
         if (isDone(c)) return
         c.onsocket(socket, addr.port, addr.host)
       }
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -219,7 +219,6 @@ async function holepunch(c, opts) {
   let { relayAddress, serverAddress, clientAddress, payload } = c.connect
 
   const remoteHolepunchable = !!(payload.holepunch && payload.holepunch.relays.length)
-
   const relayed = diffAddress(serverAddress, relayAddress)
 
   if (payload.firewall === FIREWALL.OPEN || (relayed && !remoteHolepunchable)) {
@@ -243,7 +242,9 @@ async function holepunch(c, opts) {
     return
   }
 
-  // TODO: would be better to just try local addrs in the background whilst continuing with other strategies...
+  // Race the LAN shortcut with normal holepunch when possible. A blocked LAN ping
+  // should not abort a connection that can still holepunch. If the remote is not
+  // holepunchable, the LAN ping is the only remaining path.
   if (c.lan && relayed && clientAddress.host === serverAddress.host) {
     const serverAddresses = payload.addresses4.filter(onlyNonReserved)
 
@@ -252,14 +253,19 @@ async function holepunch(c, opts) {
       const addr = Holepuncher.matchAddress(myAddresses, serverAddresses) || serverAddresses[0]
 
       const socket = c.dht.io.serverSocket
-      try {
-        await c.dht.ping(addr)
-      } catch {
-        maybeDestroyEncryptedSocket(c, HOLEPUNCH_ABORTED())
-        return
+      c.dht.ping(addr).then(onlanconnect, onlanerror)
+      if (!remoteHolepunchable) return
+
+      function onlanconnect() {
+        if (isDone(c)) return
+        c.onsocket(socket, addr.port, addr.host)
       }
-      c.onsocket(socket, addr.port, addr.host)
-      return
+
+      function onlanerror() {
+        if (!remoteHolepunchable) {
+          maybeDestroyEncryptedSocket(c, HOLEPUNCH_ABORTED())
+        }
+      }
     }
   }
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -14,7 +14,6 @@ const {
   closeRelayConnection,
   destroyRelayConnection
 } = require('./relay-connection')
-const { isPrivate } = require('bogon')
 const { ALREADY_LISTENING, NODE_DESTROYED, KEYPAIR_ALREADY_USED } = require('./errors')
 
 const HANDSHAKE_CLEAR_WAIT = 10000
@@ -423,20 +422,6 @@ module.exports = class Server extends EventEmitter {
       if (hs.relayToken === null) hs.rawStream.destroy()
     }
 
-    if (!direct && clientAddress.host === serverAddress.host) {
-      const clientAddresses = remotePayload.addresses4.filter(onlyPrivateHosts)
-
-      if (clientAddresses.length > 0 && this._shareLocalAddress) {
-        const myAddresses = await this._localAddresses()
-        const addr = Holepuncher.matchAddress(myAddresses, clientAddresses)
-
-        if (addr) {
-          hs.prepunching = setTimeout(onabort, HANDSHAKE_INITIAL_TIMEOUT)
-          return hs
-        }
-      }
-    }
-
     if (this._closing || this.suspended) return null
 
     if (ourRemoteAddr || this._neverPunch) {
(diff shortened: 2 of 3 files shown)


lib/connect.js
```diff
@@ -252,14 +253,19 @@ async function holepunch(c, opts) {
...
+
+      function onlanconnect() {
+        if (isDone(c)) return
+        c.onsocket(socket, addr.port, addr.host)
+      }
+
+      function onlanerror() {
+        if (!remoteHolepunchable) {
+          maybeDestroyEncryptedSocket(c, HOLEPUNCH_ABORTED())
+        }
```
noahlevenson, line 269: Reviewing at @marcus-pousette-hp's request... 

For context, I'm familiar with this NAT traversal IP candidate optimization problem, but not yet familiar with hyperdht's implementation. First time poking around this code.

I think this change is a clever race which AFAICT solves the problem correctly.

Only thing I wasn't sure about is the condition when holepunching is mid-flight when the LAN ping succeeds, and so `isDone(c) = false`. But I guess `c.onsocket` is the guard that resolves the race there?

Zooming out: the race paths are now pretty subtle and challenging to reason about, probably a bit too subtle for my taste.  But I suppose a different structure would trigger a much larger refactor, so this seems like a  reasonable compromise.

test/connections.js
```diff
@@ -181,6 +181,73 @@ test('createServer + connect - force holepunch', async function (t) {
...
+
+  await lc
+
+  t.ok(blockedLanPing, 'exercised the LAN ping failure path')
+
+  await server.close()
+  await a.destroy()
+  await b.destroy()
+})
+
```
noahlevenson, line 253: I think the test is correct, but I do get a little nervous whenever I see this sort of thing tested against loopback IP.  I think it's just out of domain, in the sense that the fix was designed to solve the opposite case -- the case where a NATed peer's public IP is not the same as its LAN candidate.  

(I think here `127.0.0.1` winds up in `localAddresses` which is what makes the test work, right?)

Not suggesting any changes, it was just a bit challenging to reason about whether this test was actually exercising the right paths.  Compromises for testing...

lib/connect.js
marcus-pousette-hp: https://github.com/holepunchto/hyperdht/pull/259/changes/d7db4f1deb37d9bb639b41bcaf7a8fa52420a05e 

I added a comment here, both isDone and c.onsocket correctly drop the other. The only real "issue" I would say here worth thinking about is that if the fallback (normal) Holepunch path wins, the LAN shortcut path will be ignored. In theory this would mean that we would take a little bit longer path than what the LAN shortcut path would offer

Context:
*LAN shortcut*
Uses the peer’s advertised local address, like 192.168.x.x / 10.x.x.x / loopback in the test. If both peers are truly reachable on that LAN address, this is usually the best path: fewer NAT hops, lower latency.

*Normal holepunch path*
Uses the externally observed/NAT address learned through relays/bootstrap nodes. If both peers share a public IP and the NAT supports hairpinning, this can still connect directly, but traffic may go through the router’s NAT hairpin path rather than straight peer-to-peer LAN addressing. 


This PR makes the race and the faster wins, and there is an assumption here that if LAN shortcut is the fastest path it would win anyway here, but the shortcoming of this PR is perhaps it does not prove for example a bad scenario where the normal holepunch path would win the race, even though it is slower later when data is transmitting. My own judgement finds this unlikely, which is the reason why I went with this solution

test/connections.js
marcus-pousette-hp: I see your concern. I added a clarifying comment in  https://github.com/holepunchto/hyperdht/pull/259/commits/0c06e40d5f0f032aedccbe2e28c5f35c9e019467 about the test purpose 


> (I think here 127.0.0.1 winds up in localAddresses which is what makes the test work, right?)

Yeps! In this test `127.0.0.1` is intentionally used so `Holepuncher.localAddresses(...)` returns the loopback address and we deterministically enter the same LAN shortcut branch

lib/server.js
lejeunerenard:
```suggestion
  async _addHandshake(k, noise, clientAddress, { socket }, direct) {
```
The `from` and `serverAddress` are no longer needed after removing the check for a non-direct handshake for a local client.

test/connections.js
```diff
@@ -181,6 +181,76 @@ test('createServer + connect - force holepunch', async function (t) {
   await b.destroy()
 })
 
+test('createServer + connect - failed LAN ping falls back to holepunch', async function (t) {
+  const [boot] = await swarm(t)
+
+  // Use loopback to deterministically enter the same-LAN shortcut in CI.
+  // This does not model NAT; the injected ping failure below verifies that
+  // a blocked LAN shortcut still falls through to normal holepunch.
+  const bootstrap = [{ host: '127.0.0.1', port: boot.address().port }]
```
lejeunerenard:
```suggestion
  const { bootstrap } = await swarm(t)

  // Use loopback to deterministically enter the same-LAN shortcut in CI.
  // This does not model NAT; the injected ping failure below verifies that
  // a blocked LAN shortcut still falls through to normal holepunch.
```
`swarm()` helper provides a `bootstrap`. I think the tests that use this pattern of selecting the boot strap node are older.

Also note that the bootstrap having the loopback IP doesn't affect this test as the `myAddresses` and `serverAddresses` are gotten from `a` & `b` created below. They just always have a loopback as part of the their network interfaces since that is the default host for `testnet.js`.

lejeunerenard (changes requested):
Looks good! Good to allow holepunching even when the "lan" ping doesn't work.

Only some nits and clarifications left.

test/connections.js
marcus-pousette-hp: https://github.com/holepunchto/hyperdht/pull/259/commits/df72545c1d89cfdc5c96d3c91f7be07105a5b731 fixed !

lib/server.js
marcus-pousette-hp: https://github.com/holepunchto/hyperdht/pull/259/commits/df72545c1d89cfdc5c96d3c91f7be07105a5b731 fixed !

lejeunerenard (approved):

PR #260, opened 2026-05-26
On-disk database + warm bootstrap

noahlevenson:
Adds a namespaced on-disk database and its first use case: persisting the routing table to disk, then restoring it at startup to perform a warm bootstrap.

Our new database is intended to be shared by ephemeral and persistent use cases as well as all registered plugins.

Nodes now periodically snapshot their routing table to disk on a configurable interval. 

To skip the default bootstrap and perform only a warm bootstrap, pass `warmBootstrap = true` and `bootstrap = []`.

To perform the default bootstrap and then perform a warm bootstrap, pass only `warmBootstrap = true`.


RangerMauve:
instead of dbPath, could it be db which is either a string or an instance of the db?

index.js
```diff
@@ -74,6 +97,33 @@ class HyperDHT extends DHT {
...
+
+      if (this.destroyed) return
+
+      this._routingTableSnapshotter = new Snapshotter(
+        this,
+        DB_NS.SNAPSHOTS,
+        k,
+        opts.routingTableSnapshotInterval || DEFAULTS.routingTableSnapshotInterval,
+        () => {
+          return b4a.from(JSON.stringify(this.toArray({ limit: this.table.k })))
```
lejeunerenard: TODO: Come back to this to see why it needs to be stringified first

index.js
```diff
@@ -48,6 +59,12 @@ class HyperDHT extends DHT {
     this.rawStreams = new RawStreamSet(this)
     this.plugins = new Map()
 
+    this.db = new NamespacedDB({
+      path: opts.dbPath || DB_PATH,
+      maxOpenFiles: opts.dbMaxOpenFiles || DEFAULTS.dbMaxOpenFiles,
+      optimizeFiltersForMemory: true
```
lejeunerenard: Do we need this universally or only for hypersearch? This is the first i've seen of this option actually set.

index.js
```diff
@@ -94,18 +144,37 @@ class HyperDHT extends DHT {
...
+  }
+
   async resume({ log = noop } = {}) {
     if (this._deferRandomPunch) this._lastRandomPunch = Date.now()
     await super.resume({ log })
     const resuming = []
     for (const server of this.listening) resuming.push(server.resume())
     log('Resuming hyperdht servers')
     await Promise.allSettled(resuming)
+    if (this._routingTableSnapshotter) this._routingTableSnapshotter.resume()
```
lejeunerenard: Also `await this.db.resume()`?

index.js
```diff
@@ -94,18 +144,37 @@ class HyperDHT extends DHT {
...
     log('Done, hyperdht fully resumed')
   }
 
   async suspend({ log = noop } = {}) {
     this._connectable = false // just so nothing gets connected during suspension
+    if (this._routingTableSnapshotter) this._routingTableSnapshotter.suspend()
```
lejeunerenard: Also `await this.db.suspend()`?

index.js
```diff
@@ -121,14 +190,17 @@ class HyperDHT extends DHT {
   }
 
   async destroy({ force = false } = {}) {
+    this._warmBootstrapDone()
```
lejeunerenard: TODO: Decide if the warmBoot can be marked done without awaiting the promise for it.

testnet.js
lejeunerenard: Don't think we should pass `t` like this to create the `dbPath`. Passing the test object doesn't imply to me that it should have a `dbPath` necessarily.

It can be an option passed as part of the `dhtOpts` or if we want to allow unique dirs per node, we can add a specific option for it that's an async function which is called to generate the dir during setup.

test/announces.js
```diff
@@ -226,7 +227,7 @@ test('announcer background does not over-trigger', async function (t) {
   const testnet = await swarm(t, 2) // must be <=3 (less than announcer MIN_ACTIVE) to trigger previous bug
   const bootstrap = testnet.bootstrap
 
-  const a = new HyperDHT({ bootstrap })
+  const a = new HyperDHT({ bootstrap, dbPath: await tmp(t) })
```
lejeunerenard: This test doesn't use the db right?

test/helpers/index.js
```diff
@@ -31,8 +32,9 @@ async function* spawnFixture(t, args) {
   unregisterExitHandlers()
 }
 
-function createDHT(opts) {
-  return new DHT({ ...opts, host: '127.0.0.1' })
+async function createDHT(t, opts) {
+  const dbPath = await tmp(t)
+  return new DHT({ ...opts, host: '127.0.0.1', dbPath })
```
lejeunerenard: Relating to https://github.com/holepunchto/hyperdht/pull/260/changes#r3799257425 this should also be an option IMHO.

PR #263, opened 2026-06-09
Unflake 'connection pool, server side' test on udx 1.20

HDegroote:
<strike>With UDX 1.20.6, there's a code path where the stream that got selected does not error, but closes cleanly. I think the previous test was relying on undefined behaviour, so I modified it to rely on the 'close' event instead, but I don't have full context for why ti was previously listening on 'error', so please don't merge without sanity checking that</strike>

Clarified test intent and updated the test to better reflect that instead

### DIFF
--- a/test/pool.js
+++ b/test/pool.js
@@ -32,56 +32,51 @@ test('connection pool, client side', async function (t) {
 })
 
 test('connection pool, server side', async function (t) {
+  const tConns = t.test('connections')
+  tConns.plan(4)
+  const tOpen1 = t.test('open 1')
+  tOpen1.plan(1)
+  const tOpen2 = t.test('open 2')
+  tOpen2.plan(1)
+  const tError1 = t.test('error')
+  tError1.plan(1)
+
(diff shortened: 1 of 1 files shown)


test/pool.js
```diff
@@ -47,18 +47,20 @@ test('connection pool, server side', async function (t) {
   const open = t.test('open')
   open.plan(2)
   let atLeastOneOpen = false
+  let isStream1 = false
```
HDegroote: Note: we can also listen for the close event on both streams, and `open.plan(3)`. I think that's cleaner, but not sure since I don't fully understand the previous test's intent

lejeunerenard (approved):
Great thank you! The test is clearer now being more deterministic (`socket1` opens first, etc).

The test previously relied on timing in `udx-native` to allow a socket that:
```js
socket.end()
socket.destroy(errors.DUPLICATE_CONNECTION()
```
to send the error before closing the stream. With `udx-native` update to `1.20.6` the timing changed slightly causing this implicit assumption to start failing.

PR #264, opened 2026-06-10
fix: correct error constructor reference in STREAM_NOT_CONNECTED

Kozak809:
Corrects a runtime error in `lib/errors.js` where the `STREAM_NOT_CONNECTED` static method incorrectly referenced a non-existent `STREAM_DISCONNECTED` static method for stack trace generation. This change ensures the proper method is referenced to allow correct stack trace capture.


lejeunerenard (approved):
Great! Thanks for catching that!

PR #265, opened 2026-06-13
Add Hebbian synaptic weight routing for peer connections

artiomborovinskii:
## What this does

Adds a lightweight Hebbian weight system to peer connections. Each connection tracks a weight (0-1) that increases on successful packet delivery and decreases on failure. No central coordinator, no broadcast announcements.

## The math

```
weight(t+1) = (1 - λ) × weight(t) + α × quality - β × failure
```

- `λ` (decay) — old routes fade naturally
- `α` (reward) — good connections gain weight
- `β` (penalty) — failures lose weight fast (β > α by design)

## What this gives us

**DDoS resistance.** When a node gets overloaded, it signals backpressure. Neighbors detect this instantly and route around it. No special "guardian" slot needed. The network isolates the problem organically.

**No battery drain on mobile.** A node only actively uses 2-3 routes at any time (sparse activation). The rest sit in a low-power state. No need to maintain 20 active tunnels with pings.

**Emergent backbone.** Nodes that are fast and reliable naturally accumulate higher weights. Other nodes route through them. The backbone forms itself without anyone announcing "I am a hub."

**No roles to attack.** There's no CORE or EDGE flag anywhere. An attacker can't target "hub nodes" because the topology is invisible — it only exists as local weight values that each node tracks privately.

## Changes

- **New:** `lib/synaptic-weight.js` — Hebbian update, backpressure, softmax routing
- **Modified:** `lib/connection-pool.js` — weight tracking on `ConnectionRef`, new methods for weight updates
- **Modified:** `index.js` — `SynapticWeight` integration, events, `routePacket()`
- **Modified:** `README.md` — documentation section with usage examples

## Background

This replaces rigid role assignments (SMA-style CORE/EDGE tiers) with something closer to how ant colonies or neural networks solve the same problem. Local rules, global emergence. No single point of failure, no global state, no GPU required.

The full design doc is in `live-mesh3.txt` if you want to dig into the math.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,6 +329,76 @@ Having persistent nodes in different places makes the network more decentralized
 
 For more information: [`examples/isolated-dht.mjs`](examples/isolated-dht.mjs)
 
+## Synaptic Routing (Dragon P2P)
+
+A lightweight Hebbian weight system for peer connections. Instead of rigid role assignments, each connection maintains a "synaptic weight" that increases on success and decreases on failure. The network self-organizes organically.
+
+### How it works
+
+Every peer connection gets a weight between 0 and 1. When packets flow successfully (low latency, high bandwidth), the weight goes up. When packets fail or get dropped, the weight goes down. That's it. No central coordinator, no broadcast announcements, no "I am CORE" logic.
+
+The math behind it:
+
+```
+weight(t+1) = (1 - λ) × weight(t) + α × quality - β × failure
+```
+
+Where `λ` is decay (old routes fade), `α` is reward rate, `β` is penalty rate. β is deliberately higher than α so bad connections get cut fast.
+
+### Why this matters
+
+1. **DDoS resistance.** When a node gets overloaded, it signals backpressure. Neighbors detect this instantly and route around it. No special "guardian" slot needed. The network isolates the problem organically.
+
+2. **No battery drain on mobile.** A node only actively uses 2-3 routes at any time (sparse activation). The rest sit in a low-power state. No need to maintain 20 active tunnels with pings.
+
+3. **Emergent backbone.** Nodes that are fast and reliable naturally accumulate higher weights. Other nodes route through them. The backbone forms itself without anyone announcing "I am a hub."
+
+4. **No roles to attack.** There's no CORE or EDGE flag anywhere. An attacker can't target "hub nodes" because the topology is invisible — it only exists as local weight values that each node tracks privately.
+
+### Usage
+
+The synaptic weight system is built into HyperDHT and works automatically. If you want to hook into it:
+
+```js
+const node = new DHT()
+
+// Listen for weight updates on peer connections
+node.on('peer-weight', ({ publicKey, latency, bandwidth }) => {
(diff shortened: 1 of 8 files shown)


lejeunerenard:
@artiomborovinskii thank you for being interested in and investigating using other algorithms for routing topologies.

I'm unfamiliar with Hebbian weighting and "SMA-style CORE/EDGE tiers" but looking up Baby Dragon Hatchling and Hebbian, they seem to be used for neural net optimization and haven't been able to find anything related to network routing. Regarding "SMA-style CORE/EDGE tiers" again I'm not familiar with these terms and couldn't find them via a quick search, but it sounds like there's a misunderstanding. `hyperdht` doesn't use any explicit roles or tiers for nodes. It does have the concept of bootstrap nodes, but those are only for initial connections to the DHT and could be any node.

Finally this PR doesn't seem to actually change how connections are routed. The current PR could be abstracted into a separate module and use inheritance to `extend` `HyperDHT` and the connection pool to support the same additional weights etc.

artiomborovinskii:
Hi! Following up on the feedback in #265 — you were right, and I've extracted
the idea into a standalone module instead of patching core:
**hyperdht-synaptic** (link).

What it is: a drop-in `HyperDHT` subclass that adds Hebbian weight tracking
to peer connections, fully automatically:

- each peer gets a local, private weight (0–1); handshake success/latency
  raises it, failures and stream errors lower it, and weights decay over
  time — `w(t+1) = (1-λ)w(t) + α·quality − β·failure`, with `β > α` so bad
  peers fade fast
- when `connect()` is given several `relayThrough` candidates, the relay is
  chosen by softmax over these weights instead of uniform random — so relay
  choice improves with experience, with zero protocol changes
- the engine also ships as a standalone `SynapticTracker` that accepts any
  NoiseSecretStream-like connection, so it works with hyperswarm, hypercore
  replication, or any Pear app via one line:
  `swarm.on('connection', c => tracker.watch(c))`

What it deliberately does NOT do: it doesn't touch how hyperdht establishes
or routes connections, and it adds no wire messages. It's pure local
bookkeeping + an application-level selection primitive (`choosePeer()`),
which is why it can live outside core, exactly as you suggested.

Where it helps in practice:

1. **Smarter relay choice.** Today `selectRelay()` picks uniformly at random
   from the candidates the app provides. With weights, clients converge on
   the relay that has actually been fast and reliable *for them*, and route
   around a degraded relay within a few failures — no coordination, no
   announcements, invisible to attackers (weights are local-only).
2. **Peer selection for apps.** Anything built on hyperswarm that fans out
   requests (data sync, gossip, block requests) can call `choosePeer()` to
   prefer historically good peers while `λ`-decay keeps exploring — an
   ACO-style exploit/explore balance in ~50 lines of algebra, no ML, no
   extra CPU or battery cost (updates are O(1) additions/multiplications on
   existing events).
3. **Graceful degradation.** `setBackpressure(key, true)` lets an app that
   detects overload penalize a peer's weight so traffic drains away from it
   organically, instead of hammering it until sockets die.

I'd love feedback on whether the relay-selection hook (`relayThrough` as a
weighted pick) is something you'd consider useful upstream; everything else
stays in the module. Closing #265 in favor of this.

PR #266, opened 2026-06-15
Fix relayed keepalive and direct-upgrade handoff

marcus-pousette-hp:
This fixes server-side relayed connections so they inherit `connectionKeepAlive` the same way direct server sockets and client sockets already do.

Enabling that exposed a relay-to-direct upgrade race: HyperDHT was closing the relay transport after local `remote-changed`, but that only proves this side switched to direct. The peer may not have observed the direct path yet. To avoid stranding the peer, the upgrade path now sends one zero-length UDX message after `changeRemote()` so the other side can observe the new direct path before relay cleanup.

One subtle behavior change here is that the old server-side relayed path had `keepAlive: 0`, so empty secret-stream keepalive frames from the client could reach app code. In the upgrade test, the server echoes all `data`, so those empty frames could accidentally create traffic during the relay-to-direct handoff. Once server-side keepalive is enabled, secret-stream correctly consumes those empty frames instead, so the upgrade path needs its own explicit UDX-level handoff packet.

Co-authored with AI

### DIFF
--- a/test/relaying.js
+++ b/test/relaying.js
@@ -273,11 +273,16 @@ test('relay connections through node, server side', async function (t) {
   const { bootstrap } = await swarm(t)
 
   const a = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
-  const b = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
+  const b = createDHT({
+    bootstrap,
+    quickFirewall: false,
+    ephemeral: true,
+    connectionKeepAlive: 12345
+  })
   const c = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
 
   const lc = t.test('socket lifecycle')
-  lc.plan(5)
+  lc.plan(6)
 
   const relay = new RelayServer({
     createStream(opts) {
@@ -296,6 +301,7 @@ test('relay connections through node, server side', async function (t) {
 
   const bServer = b.createServer({ relayThrough: aServer.publicKey }, function (socket) {
     lc.pass('server socket opened')
+    lc.is(socket.keepAlive, 12345, 'server relayed socket inherits connectionKeepAlive')
     socket
       .on('data', (data) => {
         lc.alike(data, Buffer.from('hello world'))
--- a/test/relaying.js
+++ b/test/relaying.js
@@ -542,115 +542,149 @@ test('relay connections through node, client and server side', async function (t
 })
 
 test('relay connection upgrades to direct connection', async function (t) {
-  const { bootstrap } = await swarm(t)
+  for (const opts of [
+    { name: 'default keepalive' },
+    { name: 'without keepalive', connectionKeepAlive: false, confirmWithAppData: true },
+    { name: 'idle without keepalive', connectionKeepAlive: false }
+  ]) {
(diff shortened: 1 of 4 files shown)


test/relaying.js
```diff
@@ -296,6 +301,7 @@ test('relay connections through node, server side', async function (t) {
 
   const bServer = b.createServer({ relayThrough: aServer.publicKey }, function (socket) {
     lc.pass('server socket opened')
+    lc.is(socket.keepAlive, 12345, 'server relayed socket inherits connectionKeepAlive')
```
lejeunerenard: This assertion passes without the `keepAlive` change in `server.js` so isn't covering that case.

lejeunerenard (commented):
Addressed/will address the upgrading of the relay connection in the other stacked PRs. Looking at the connectionKeepAlive change and found the new assertion isn't testing the missing configuration path.

test/relaying.js
marcus-pousette-hp: Fixed! https://github.com/holepunchto/hyperdht/pull/266/commits/9dc78333ee47242028d204ce2a5f35370e1e3463

test/relaying.js
```diff
@@ -486,7 +495,19 @@ test('relay connections through node, client and server side', async function (t
     {
       holepunch: false, // To ensure it relies only on relaying
       shareLocalAddress: false, // To help ensure it relies only on relaying (otherwise it can connect directly over LAN, without even trying to holepunch)
-      relayThrough: aServer.publicKey
+      relayThrough: aServer.publicKey,
+      createSecretStream(isInitiator, rawStream, opts) {
+        if (!isInitiator && !assertedRelayKeepAlive) {
```
lejeunerenard: Is the `assertedRelayKeepAlive` necessary? I've tried removing it and the assert is only ever called once.

lejeunerenard (commented):
Just one nit: Otherwise looks good to merge #268 in.

test/relaying.js
marcus-pousette-hp: good point, missed it

marcus-pousette-hp: Fixed https://github.com/holepunchto/hyperdht/pull/266/commits/f19226fb0a73d0ed02bf413e20dd0d7cb19a8725

marcus-pousette-hp:
~So I did an investigation about validUpgrade flag flipflop and we seem to have bug~

~https://github.com/holepunchto/hyperdht/pull/271 (stacked PR with just the test)~

~The bug is that HyperDHT is trying to decide whether relay-to-direct upgrade is confirmed from raw data/message events, but those events do not say whether the packet came from the old relay socket or the new direct socket.~

~In theory in-flight messages could screw things up with this approach but is fixable if we patch libudx/udx-native fix we can fix this cleanly but require more work.~

I amend my previous comment (above) as invalid conclusions and I reminded myself of the implied order assumptions we do around the firewall callback. 

I would argue there is missing test coverage for the order of the callbacks: https://github.com/holepunchto/libudx/pull/293 this would prevent a downstream refactor to potentially screw with assumption we are making with this PR. 


Otherwise this PR is good to go now!

lejeunerenard (approved):

PR #267, opened 2026-06-15
Sort results to check response w/ most peers for assertions

lejeunerenard:
The test can announce two slightly different sets of peers from various of response timings causing the flake.

Having the exact same set of peers responding to a target (even from the same peer) is not the nature of the test nor does it reflect real world use case where network timings result in lots of variation. So the result is sorted so the peer selected is more likely to have both announcements to stabilize the test.

### DIFF
--- a/test/announces.js
+++ b/test/announces.js
@@ -52,13 +52,20 @@ test('announce to group and lookup', async function (t) {
   {
     const result = await toArray(b.lookup(target))
     t.ok(result.length > 0, 'has at least one result')
-    t.alike(result[0].peers.length, 2, 'two peers')
+
+    // Sort result by number of peers to first is more likely to have both announcements
+    // The test can return different set of nodes per lookup for announcing so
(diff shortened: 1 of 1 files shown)


test/announces.js
```diff
@@ -52,13 +52,21 @@ test('announce to group and lookup', async function (t) {
   {
     const result = await toArray(b.lookup(target))
     t.ok(result.length > 0, 'has at least one result')
-    t.alike(result[0].peers.length, 2, 'two peers')
+
```
HDegroote, line 55: I think you can also solve by creating a swarm with fewer nodes (on line 36), maybe just 2? Default is 32

HDegroote: Note: feel free to resolve, we discussed out of band that this was already considered

HDegroote (approved):
Makes sense to me now (my earlier review had misunderstood the test)

PR #268, opened 2026-06-17
Validate upgrading to direct when changing remote from relay to direct

lejeunerenard:
Changing the remote and immediately closing the relay connection leads to situations where one side is left trying to ack the packets sent via the relay connection but unable to since the peer closed their side since the remote changed successfully.

Instead this approach uses the fact that changing the remote (whether immediately or deferred) changes the stream's socket which means that the `firewall` callback will be called for each packet received that comes from a different socket, in the case of changing remote from the relay -> direct means `firewall` only triggers from the relay.

This means the direct connection can be validated by waiting for a packet to come through that doesn't trigger the firewall. This is done by setting a flag on the connection / hs that assumes the next `data` event will be direct. The flag is set to false when the firewall is called and the socket its called with is shown to be the relay's. Once direct the relay connection is cleaned up via the normal method.

Collaborated with AI to debug.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -11,7 +11,7 @@ const Holepuncher = require('./holepuncher')
 const Sleeper = require('./sleeper')
 const {
   clearRelayTimeout,
-  closeRelayConnection,
+  confirmDirectUpgrade,
   destroyRelayConnection
 } = require('./relay-connection')
 const { FIREWALL, ERROR } = require('./constants')
@@ -500,28 +500,9 @@ async function connectThroughNode(c, address, socket) {
 
     if (c.rawStream.connected) {
       c.validUpgrade = true
-      const remoteChanging = c.rawStream.changeRemote(socket, c.connect.payload.udx.id, port, host)
-      const confirmDirect = () => {
-        if (c.validUpgrade) {
-          closeRelayConnection(c)
-          if (c.rawStream) c.rawStream.off('data', confirmDirect)
-        }
-        c.validUpgrade = true // reset, aka assume from direct
-      }
-
-      if (remoteChanging) {
-        remoteChanging.then(() => {
-          if (c.rawStream) {
-            c.rawStream.on('data', confirmDirect)
-          } else {
-            // TODO verify this gets hit, probably via queue packets before remote change
-            closeRelayConnection(c)
-          }
-        })
-        remoteChanging.catch(safetyCatch)
-      } else {
-        c.rawStream.on('data', confirmDirect)
-      }
+      const rawStream = c.rawStream
+      const remoteChanging = rawStream.changeRemote(socket, c.connect.payload.udx.id, port, host)
+      confirmDirectUpgrade(c, rawStream, remoteChanging)
     } else {
       // cache the relay addrs for a future reconnect, we prefer the remote one so they
       // can give us the correct ones from their pov
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -102,6 +102,7 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     relaySocket: null,
     relayClient: null,
     relayPaired: false,
+    validUpgrade: true,
     relayKeepAlive: opts.relayKeepAlive || 5000
   }
 
@@ -135,9 +136,11 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     // Check if the traffic originated from the socket on which we're expecting relay traffic. If so,
     // we haven't hole punched yet and the other side is just sending us traffic through the relay.
     if (c.relaySocket && isRelay(c.relaySocket, socket, port, host)) {
+      c.validUpgrade = false
       return false
     }
 
+    c.validUpgrade = true
     if (c.onsocket) {
       c.onsocket(socket, port, host)
     } else {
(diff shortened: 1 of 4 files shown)


lib/connect.js
```diff
@@ -135,9 +136,11 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     // Check if the traffic originated from the socket on which we're expecting relay traffic. If so,
     // we haven't hole punched yet and the other side is just sending us traffic through the relay.
     if (c.relaySocket && isRelay(c.relaySocket, socket, port, host)) {
+      c.validUpgrade = false
```
marcus-pousette-hp, line 139: I am wodnering here if we can have a scenario where we first get a direct packet and validUpgrade is set to be true, and then we get a relay packet and it gets set to be false again, just because there is communication happening on both. And this would prevent the handoff from correctly or efficiently happening

lib/server.js
```diff
@@ -291,10 +292,12 @@ module.exports = class Server extends EventEmitter {
           // Check if the traffic originated from the socket on which we're expecting relay traffic. If so,
           // we haven't hole punched yet and the other side is just sending us traffic through the relay.
           if (hs.relaySocket && isRelay(hs.relaySocket, socket, port, host)) {
+            hs.validUpgrade = false
```
marcus-pousette-hp, line 295: same as the other comment here about the validUpgrade flipping back and forth

marcus-pousette-hp (commented):
What would be also nice to have is a  
- Relay to direct upgrade when connectionKeepAlive: false
- Coverage for what happens when late relay packets arrive and so that bugs like the validUpgrade flag is flippening back/forth is not present

lib/connect.js
lejeunerenard: Packets received should be strictly in order I believe because the peer should only send direct packets when the stream successfully switches over sockets. That can only happen when the relayed packets have all been acked:
https://github.com/holepunchto/libudx/blob/759bf7641347f65a1896d4bc4c3b1606a459f466/src/udx.c#L1593

lejeunerenard:
> Relay to direct upgrade when connectionKeepAlive: false
@marcus-pousette-hp currently this is still possible, the relay connection will just not be closed until a packet is received from the direct connection as we cannot know if the other peer has `connectionKeepAlive` set to `false` too.

Since we account for when the remote is [changed immediately](https://github.com/holepunchto/hyperdht/pull/268/changes#diff-459561ac30417a9e40a7904e9a3c297a6c840189bf21583d951746a1d43a506aR523) (aka immediate direct connection upgrade), it doesnt rely on packets to upgrade locally, but relies on packets from the peer to verify it can close the relay connection.

lib/connect.js
marcus-pousette-hp: You are right, this make my point invalid!

marcus-pousette-hp:
Co-edited with AI below, 

For me the first test passes for me below, but the second does not

```ts
  await testRelayUpgrade(t)
})

test('relay connection upgrades to direct connection without keepalive', async function (t) {
  await testRelayUpgrade(t, { connectionKeepAlive: false })
})
```

where

```ts
async function testRelayUpgrade(t, opts = {}) {

  const dhtOpts =
    opts.connectionKeepAlive === false ? { connectionKeepAlive: false } : {}

  const relayNode = createDHT({ bootstrap, ...dhtOpts })
  const serverNode = createDHT({
    bootstrap,
    quickFirewall: false,
    ephemeral: true,
    ...dhtOpts
  })
  const clientNode = createDHT({
    bootstrap,
    quickFirewall: false,
    ephemeral: true,
    ...dhtOpts
  })

  const resumePunching = pausePunching(t, [serverNode, clientNode])

  const relayServer = new RelayServer({
    createStream(opts) {
      return relayNode.createRawStream({ ...opts, framed: true })
    }
  })

  t.teardown(() => relayServer.close())

  const relaySockets = []
  let resolveRelaySockets = null
  const relaySocketsOpened = new Promise((resolve) => {
    resolveRelaySockets = resolve
  })

  const relayTransportServer = relayNode.createServer(function (socket) {
    relaySockets.push(socket)
    if (relaySockets.length === 2) resolveRelaySockets(relaySockets)

    const session = relayServer.accept(socket, { id: socket.remotePublicKey })
    session.on('error', (err) => t.comment(err.message))
  })

  await relayTransportServer.listen()

  let resolveServerSocket = null
  const serverSocketOpened = new Promise((resolve) => {
    resolveServerSocket = resolve
  })

  const appServer = serverNode.createServer(
    {
      relayThrough: relayTransportServer.publicKey,
      shareLocalAddress: false
    },
    function (socket) {
      resolveServerSocket(socket)

      socket.on('data', (data) => socket.write(data))
      socket.on('end', () => socket.end())
    }
  )

  await appServer.listen()

  const clientSocket = clientNode.connect(appServer.publicKey, {
    fastOpen: false,
    localConnection: false
  })

  const [serverSocket] = await Promise.all([
    serverSocketOpened,
    once(clientSocket, 'open')
  ])
  await relaySocketsOpened

  t.is(relaySockets.length, 2, 'both peers opened relay transport sockets')

  t.not(
    serverSocket.rawStream.remotePort,
    clientSocket.rawStream.localPort,
    'server starts on the relayed stream'
  )
  t.not(
    clientSocket.rawStream.remotePort,
    serverSocket.rawStream.localPort,
    'client starts on the relayed stream'
  )

  const beforeUpgrade = once(clientSocket, 'data')
  clientSocket.write(Buffer.from('before upgrade'))
  t.alike(
    (await beforeUpgrade)[0],
    Buffer.from('before upgrade'),
    'relay path carries data'
  )

  const clientUpgraded = once(clientSocket.rawStream, 'remote-changed')
  const serverUpgraded = once(serverSocket.rawStream, 'remote-changed')
  const relaySocketsClosed = relaySockets.map((socket) => once(socket, 'close'))

  resumePunching()

// here is the critical point
  await Promise.all([clientUpgraded, serverUpgraded])
  await Promise.all(relaySocketsClosed)

  t.pass('relay transport sockets close after direct upgrade')

  t.is(
    serverSocket.rawStream.remotePort,
    clientSocket.rawStream.localPort,
    'server switches to the client address'
  )
  t.is(
    clientSocket.rawStream.remotePort,
    serverSocket.rawStream.localPort,
    'client switches to the server address'
  )

  const afterUpgrade = once(clientSocket, 'data')
  clientSocket.write(Buffer.from('after upgrade'))
  t.alike(
    (await afterUpgrade)[0],
    Buffer.from('after upgrade'),
    'direct path carries data'
  )

  await endAndCloseSocket(clientSocket)
  if (!serverSocket.destroyed) await once(serverSocket, 'close')

  await relayNode.destroy()
  await serverNode.destroy()
  await clientNode.destroy()
}
```

PR #269, opened 2026-06-18
Test/relay upgrade passive confirmation

marcus-pousette-hp:
Adds test coverage to lock down the assumptions made in #268    

Co-authored with AI

### DIFF
--- a/test/relaying.js
+++ b/test/relaying.js
@@ -528,151 +528,149 @@ test('relay connections through node, client and server side', async function (t
 })
 
 test('relay connection upgrades to direct connection', async function (t) {
-  await testRelayUpgrade(t)
-})
+  for (const opts of [
+    { name: 'default keepalive' },
+    { name: 'without keepalive', connectionKeepAlive: false, confirmWithAppData: true }
+  ]) {
+    t.comment(opts.name)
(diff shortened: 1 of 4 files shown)


test/relaying.js
```diff
@@ -528,11 +528,29 @@ test('relay connections through node, client and server side', async function (t
 })
 
 test('relay connection upgrades to direct connection', async function (t) {
+  await testRelayUpgrade(t)
+})
+
+test('relay connection upgrades to direct connection without keepalive', async function (t) {
+  await testRelayUpgrade(t, { connectionKeepAlive: false, confirmWithAppData: true })
+})
```
lejeunerenard: nit: I think i'd rather this was moved into a single test that runs twice via a loop with the params as what it is looping. It represents it better in that it directly asserts in the test (instead of a helper containing the entire test) and still shows that the assertions should work on two different configurations.

marcus-pousette-hp: Ye makes sense! Fixed in https://github.com/holepunchto/hyperdht/pull/269/commits/c5bf494a0563e10c60f50f20a684ed1d5f466232

lejeunerenard (approved):

PR #270, opened 2026-06-18
fix: nudge idle relay upgrades

marcus-pousette-hp:
A an attempt to make the nudging part of the intended behaviour when connectionKeepAlive is false

Co-authored with AI

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -11,7 +11,7 @@ const Holepuncher = require('./holepuncher')
 const Sleeper = require('./sleeper')
 const {
   clearRelayTimeout,
-  closeRelayConnection,
+  confirmDirectUpgrade,
   destroyRelayConnection
 } = require('./relay-connection')
 const { FIREWALL, ERROR } = require('./constants')
@@ -103,6 +103,7 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     relayClient: null,
     relayPaired: false,
     validUpgrade: true,
+    directUpgradeObserved: false,
     relayKeepAlive: opts.relayKeepAlive || 5000
   }
 
@@ -136,10 +137,12 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     // Check if the traffic originated from the socket on which we're expecting relay traffic. If so,
     // we haven't hole punched yet and the other side is just sending us traffic through the relay.
     if (c.relaySocket && isRelay(c.relaySocket, socket, port, host)) {
+      c.directUpgradeObserved = false
       c.validUpgrade = false
       return false
     }
 
+    c.directUpgradeObserved = true
     c.validUpgrade = true
     if (c.onsocket) {
       c.onsocket(socket, port, host)
@@ -500,28 +503,17 @@ async function connectThroughNode(c, address, socket) {
 
     if (c.rawStream.connected) {
       c.validUpgrade = true
-      const remoteChanging = c.rawStream.changeRemote(socket, c.connect.payload.udx.id, port, host)
-      const confirmDirect = () => {
-        if (c.validUpgrade) {
-          closeRelayConnection(c)
-          if (c.rawStream) c.rawStream.off('data', confirmDirect)
-        }
-        c.validUpgrade = true // reset, aka assume from direct
-      }
+      const directObserved = c.directUpgradeObserved === true
+      c.directUpgradeObserved = false
+      const rawStream = c.rawStream
+      const remoteChanging = rawStream.changeRemote(socket, c.connect.payload.udx.id, port, host)
+      const confirming = confirmDirectUpgrade(c, rawStream, remoteChanging, {
+        directObserved,
+        confirmNudge: c.dht.connectionKeepAlive === 0,
+        nudge: c.dht.connectionKeepAlive === 0
+      })
 
-      if (remoteChanging) {
-        remoteChanging.then(() => {
-          if (c.rawStream) {
-            c.rawStream.on('data', confirmDirect)
-          } else {
-            // TODO verify this gets hit, probably via queue packets before remote change
-            closeRelayConnection(c)
-          }
-        })
-        remoteChanging.catch(safetyCatch)
-      } else {
-        c.rawStream.on('data', confirmDirect)
-      }
+      if (confirming) confirming.catch(safetyCatch)
     } else {
       // cache the relay addrs for a future reconnect, we prefer the remote one so they
       // can give us the correct ones from their pov
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -103,7 +103,6 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     relayClient: null,
     relayPaired: false,
     validUpgrade: true,
-    directUpgradeObserved: false,
     relayKeepAlive: opts.relayKeepAlive || 5000
   }
 
@@ -137,12 +136,10 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     // Check if the traffic originated from the socket on which we're expecting relay traffic. If so,
     // we haven't hole punched yet and the other side is just sending us traffic through the relay.
     if (c.relaySocket && isRelay(c.relaySocket, socket, port, host)) {
-      c.directUpgradeObserved = false
       c.validUpgrade = false
       return false
     }
 
-    c.directUpgradeObserved = true
     c.validUpgrade = true
     if (c.onsocket) {
       c.onsocket(socket, port, host)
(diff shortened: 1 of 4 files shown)


lib/relay-connection.js
```diff
@@ -13,6 +16,61 @@ function closeRelayConnection(connectionOrHandshake) {
   if (socket) socket.end()
 }
 
+function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging, opts = {}) {
+  let remoteChanged = !remoteChanging
+  let directObserved = opts.directObserved === true
+  let nudged = false
+  let done = false
```
lejeunerenard: The `done` guard isn't needed since `close()` can't be executed after it's run once.

lib/relay-connection.js
```diff
@@ -13,6 +16,61 @@ function closeRelayConnection(connectionOrHandshake) {
   if (socket) socket.end()
 }
 
+function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging, opts = {}) {
+  let remoteChanged = !remoteChanging
+  let directObserved = opts.directObserved === true
+  let nudged = false
```
lejeunerenard: The `nudged` guard isn't needed, right? I can only be run immediately if the remote changed immediately or after the `remoteChanging` promise resolves so its mutually exclusive, right?

lib/connect.js
```diff
@@ -103,6 +103,7 @@ module.exports = function connect(dht, publicKey, opts = {}) {
     relayClient: null,
     relayPaired: false,
     validUpgrade: true,
+    directUpgradeObserved: false,
```
lejeunerenard: I dont understand what this flag gives us that isn't already covered by `validUpgrade`. It is only assigned besides the check that it's true in the relaying path of the `onsocket` callback which will never be true because its checked in the path it takes to start the upgrade.

lib/relay-connection.js
```diff
@@ -13,6 +16,61 @@ function closeRelayConnection(connectionOrHandshake) {
   if (socket) socket.end()
 }
 
+function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging, opts = {}) {
+  let remoteChanged = !remoteChanging
+  let directObserved = opts.directObserved === true
```
lejeunerenard: If `directUpgradeObserved` is always `false` as mentioned above, then this can be defaulted to `false`.

lib/relay-connection.js
```diff
@@ -13,6 +16,61 @@ function closeRelayConnection(connectionOrHandshake) {
...
+  rawStream.once('close', cleanup)
+
+  const nudge = () => {
+    if (nudged) return
+    nudged = true
+
+    // Use a raw UDX message to make the peer observe the direct path without
+    // writing application data into the secret stream.
+    if (opts.nudge && rawStream.trySend) rawStream.trySend(b4a.alloc(0))
+    if (opts.confirmNudge) directObserved = true
```
lejeunerenard: This is artificial right? The direct path isn't observed yet. There's no guarantee that the nudge packet makes it to the server and we'd want to actually observe the server using the direct socket as our indication to close the relay connection.

lib/relay-connection.js
```diff
@@ -13,6 +16,61 @@ function closeRelayConnection(connectionOrHandshake) {
   if (socket) socket.end()
 }
 
+function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging, opts = {}) {
+  let remoteChanged = !remoteChanging
```
lejeunerenard: `remoteChanged` isn't necessary as we only need to know if the remote changed immediately (aka `remoteChanging === null`) to skip awaiting it. All the checks for `remoteChanged` are in places that we know that the remote changed either because its awaiting `remoteChanging` or via a packet on the direct connection.

lejeunerenard (commented):
I think moving the confirmation / validation of the direct connection into a single function to reuse in both the client & the server is a good idea.

There seems like a lot of extra guards in the current implementation that I'm not sure that we need.

lib/relay-connection.js
marcus-pousette-hp: Ye that is true, prior solution was overly defensive. Fixed #https://github.com/holepunchto/hyperdht/pull/270/commits/e425786deac7c1f444ad4057e68aefe7814be308

lib/relay-connection.js
marcus-pousette-hp: Agreed, removed. After simplifying the flow, the nudge only runs when confirmation starts https://github.com/holepunchto/hyperdht/commit/e425786deac7c1f444ad4057e68aefe7814be308

lib/connect.js
marcus-pousette-hp: Yeah, not necessary. Did not pay enough attention here. Fixed https://github.com/holepunchto/hyperdht/pull/270/commits/e425786deac7c1f444ad4057e68aefe7814be308

lib/relay-connection.js
marcus-pousette-hp: Yeah, removed it entirely now [e425786](https://github.com/holepunchto/hyperdht/pull/270/commits/e425786deac7c1f444ad4057e68aefe7814be308)

lib/connect.js
marcus-pousette-hp: Originally it was meant to carry the firewall callback "this packet was not relay traffic"

lib/relay-connection.js
marcus-pousette-hp: Agreed. That was too optimistic. It should not count as confirmation itself https://github.com/holepunchto/hyperdht/pull/270/commits/e425786deac7c1f444ad4057e68aefe7814be308

lib/relay-connection.js
marcus-pousette-hp: Yeah, agreed. Since we can start confirmation only after changeRemote() is ready, the helper no longer needs to track remoteChanged state. https://github.com/holepunchto/hyperdht/pull/270/commits/e425786deac7c1f444ad4057e68aefe7814be308

lib/connect.js
```diff
@@ -500,28 +500,13 @@ async function connectThroughNode(c, address, socket) {
...
-          } else {
-            // TODO verify this gets hit, probably via queue packets before remote change
-            closeRelayConnection(c)
-          }
-        })
-        remoteChanging.catch(safetyCatch)
-      } else {
-        c.rawStream.on('data', confirmDirect)
-      }
+      if (confirming) confirming.catch(safetyCatch)
```
lejeunerenard: Should we just move this into `confirmDirectUpgrade`? Given all the logic around the upgrade is in there catching when it's a promise (which is also checked there already) makes it easier on the caller.

lib/relay-connection.js
```diff
@@ -13,6 +16,41 @@ function closeRelayConnection(connectionOrHandshake) {
...
+function confirmDirectUpgrade(connectionOrHandshake, rawStream, remoteChanging, opts = {}) {
+  const cleanup = () => {
+    rawStream.off('data', ondirect)
+    rawStream.off('message', ondirect)
+    rawStream.off('close', cleanup)
+  }
+
+  function ondirect() {
+    if (!connectionOrHandshake.validUpgrade) {
+      connectionOrHandshake.validUpgrade = true
```
lejeunerenard, line 30:
```suggestion
      // reset, aka assume from direct
      connectionOrHandshake.validUpgrade = true
```
Let's add back the comment from before refactoring to explain. Otherwise this could be confusing in the future.

lib/relay-connection.js
```diff
@@ -13,6 +16,41 @@ function closeRelayConnection(connectionOrHandshake) {
...
+  }
+
+  const confirm = () => {
+    rawStream.on('data', ondirect)
+    rawStream.on('message', ondirect)
+    rawStream.once('close', cleanup)
+
+    // Use a raw UDX message to make the peer observe the direct path without
+    // writing application data into the secret stream.
+    if (opts.nudge && rawStream.trySend) rawStream.trySend(b4a.alloc(0))
```
lejeunerenard: We should move the logic for when to `nudge` into here if we keep it (more on that below). Specifically we should check the `keepAlive` on the `encryptedSocket` instead of the option in the `dht`. Since its the encryptedSocket that would be otherwise sending the keep alive packet that the remote would observe instead of the nudge packet, we should check it just to be sure it wasn't changed later etc.

That all doesn't matter however if we decide to always nudge the remote. I think this makes sense because then the remote will not continue holepunching for the duration of the keep alive timeout in the scenario where no app data is sent before the keep alive. The current keep alive is 5s so why make the remote keep holepunching when we know that we already succeeded. Sending an extra packet is not ideal, but it only happens when upgrading from a relay. So its minimal.

marcus-pousette-hp: Agree with you.. always nudging is simpler and I think we should go with that. Fixed in https://github.com/holepunchto/hyperdht/pull/270/commits/1a62a66c19215ac1ab376db1371df06c4016a93d

lib/relay-connection.js
marcus-pousette-hp: Fixed https://github.com/holepunchto/hyperdht/pull/270/commits/1a62a66c19215ac1ab376db1371df06c4016a93d

lib/connect.js
marcus-pousette-hp: Yeah, makes it easier. Fixed https://github.com/holepunchto/hyperdht/pull/270/commits/1a62a66c19215ac1ab376db1371df06c4016a93d

lejeunerenard (approved):

PR #271, opened 2026-06-26
test: expose relay upgrade confirmation race

marcus-pousette-hp:
co-written with AI


marcus-pousette-hp:
Closing as this does not seem to be an issue, and the initial test was wrongly defined to showcase an issue with in-flight  relay messages delivered later

PR #272, opened 2026-06-29
Fix server lan prepunch

marcus-pousette-hp:
This PR restores the server-side same-LAN prepunch path (https://github.com/holepunchto/hyperdht/blob/a9371f4142a49b055e73dbe952c7716b8bffab34/lib/server.js#L426-L436)  so explicit-keypair connections can open on both peers again, while preserving the existing LAN-ping fallback to normal holepunching.

Co-written with AI

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -431,7 +431,7 @@ module.exports = class Server extends EventEmitter {
     }
 
     if (!direct && clientAddress.host === serverAddress.host) {
-      const clientAddresses = remotePayload.addresses4.filter(onlyPrivateHosts)
+      const clientAddresses = remotePayload.addresses4.filter((addr) => isPrivate(addr.host))
 
       if (clientAddresses.length > 0 && this._shareLocalAddress) {
         const myAddresses = await this._localAddresses()
@@ -728,10 +728,6 @@ function setupHolepuncher(dht, hs, remoteFirewall) {
   hs.puncher = new Holepuncher(dht, dht.session(), false, remoteFirewall)
   hs.puncher.onconnect = hs.onsocket
   hs.puncher.onabort = hs.onabort
-}
-
-function onlyPrivateHosts(addr) {
-  return isPrivate(addr.host)
 }
 
 function isRelay(relaySocket, socket, port, host) {
--- a/lib/server.js
+++ b/lib/server.js
@@ -14,6 +14,7 @@ const {
   closeRelayConnection,
   destroyRelayConnection
 } = require('./relay-connection')
+const { isPrivate } = require('bogon')
 const { ALREADY_LISTENING, NODE_DESTROYED, KEYPAIR_ALREADY_USED } = require('./errors')
 
 const HANDSHAKE_CLEAR_WAIT = 10000
@@ -213,7 +214,7 @@ module.exports = class Server extends EventEmitter {
     return this.dht.validateLocalAddresses(Holepuncher.localAddresses(this.dht.io.serverSocket))
   }
 
-  async _addHandshake(k, noise, clientAddress, { socket }, direct) {
+  async _addHandshake(k, noise, clientAddress, { to: serverAddress, socket }, direct) {
     let id = this._holepunches.indexOf(null)
     if (id === -1) id = this._holepunches.push(null) - 1
 
@@ -227,6 +228,7 @@ module.exports = class Server extends EventEmitter {
       prepunching: null,
       firewalled: true,
       clearing: null,
+      onabort: null,
       onsocket: null,
       aborted: false,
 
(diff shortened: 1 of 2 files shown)


lib/server.js
```diff
@@ -430,10 +453,7 @@ module.exports = class Server extends EventEmitter {
     }
 
     hs.payload = new SecurePayload(h.holepunchSecret)
-    hs.puncher = new Holepuncher(this.dht, this.dht.session(), false, remotePayload.firewall)
-
-    hs.puncher.onconnect = hs.onsocket
-    hs.puncher.onabort = onabort
+    ensureHolepuncher()
```
lejeunerenard: Should just call `setupHolepuncher` instead of creating a closures for one call.

lib/server.js
```diff
@@ -482,13 +502,16 @@ module.exports = class Server extends EventEmitter {
...
     if (!peerAddress || this._closing !== null || this.suspended) return null
-
-    const p = h.puncher
-    if (!p || !p.socket) return this._abort(h) // not opened
+    if (!h.payload) return this._abort(h) // not opened
 
     const remotePayload = h.payload.decrypt(payload)
     if (!remotePayload) return null
 
+    setupHolepuncher(this.dht, h, remotePayload.firewall)
```
lejeunerenard: We should move this to `_addHandshake()` so holepunching starts after the handshake is finalized. Allowing to start holepunching after receiving the `PEER_HOLEPUNCH` message is fine but deviates from the path every non-LAN connection takes.

lejeunerenard (changes requested):
These changes now use a different pathway for starting holepunching on the server side. Not an issue per se, but having the same path as the other peer types is simpler.

lib/server.js
marcus-pousette-hp: Good point. Fixed https://github.com/holepunchto/hyperdht/pull/272/commits/45aade615bbb4b6d7562088b6e0905fdd710ef17

lib/server.js
marcus-pousette-hp: Fixed https://github.com/holepunchto/hyperdht/pull/272/commits/45aade615bbb4b6d7562088b6e0905fdd710ef17

lib/server.js
```diff
@@ -474,21 +497,22 @@ module.exports = class Server extends EventEmitter {
 
     if (this._closing !== null || this.suspended) return null
 
-    return { socket: h.puncher && h.puncher.socket, noise: h.reply }
+    return { socket: h.handshakeSocket, noise: h.reply }
```
lejeunerenard: Why cache and then use the socket instead of getting it from the puncher like before? The socket can change between then and now.

lejeunerenard (changes requested):

lib/server.js
marcus-pousette-hp: Good point. I cached the socket to separate "same-LAN should reply on the default request socket" from "normal holepunch should reply on the puncher socket", but that accidentally made the normal path snapshot the puncher socket wrongly. 

I am comitting a fix https://github.com/holepunchto/hyperdht/pull/272/commits/413e2151230fa066d89dce2faff7a15908c86177 , but without test coverage. In practice we could in theory do a test coverage here even but it would not look very nice (but could be done if we monkey patch _addHandshake in a test)

lib/server.js
```diff
@@ -422,6 +429,24 @@ module.exports = class Server extends EventEmitter {
       if (hs.relayToken === null) hs.rawStream.destroy()
     }
 
+    hs.onabort = onabort
```
lejeunerenard: We can probably get rid of this and assign this directly to the holepuncher instead. No need to keep the property on the `hs` now that the holepuncher is defined here.

marcus-pousette-hp: Ye, I think it looks better if we do that way. Fixed https://github.com/holepunchto/hyperdht/pull/272/commits/36dbc5aa2c7584ac944b171845c2c51dadec200d

marcus-pousette-hp:
After this PR, same-LAN prepunch also creates `h.puncher`, but that puncher is only a helper socket. If `_onpeerhandshake()` returns that socket, the routed handshake reply advertises the helper socket as the server address in [`lib/router.js#L162-L166`](https://github.com/holepunchto/hyperdht/blob/ce02cb2/lib/router.js#L162-L166).

Then the client can mark the puncher connected in [`lib/holepuncher.js#L124-L145`](https://github.com/holepunchto/hyperdht/blob/ce02cb2/lib/holepuncher.js#L124-L145). Once `c.puncher.connected` is true, [`isDone(c)` returns true](https://github.com/holepunchto/hyperdht/blob/ce02cb2/lib/connect.js#L151-L154), so when the LAN shortcut later succeeds it returns before `c.onsocket(...)` in [`lib/connect.js#L259-L263`](https://github.com/holepunchto/hyperdht/blob/ce02cb2/lib/connect.js#L259-L263).

That means the client-side puncher can “win” against the helper socket before the server stream is opened, and then the actual same-LAN path is skipped.


Added some clarification here https://github.com/holepunchto/hyperdht/pull/272/commits/ce02cb2a8917f9eb7cca498bee7db7c8e8f95540

lejeunerenard (approved):

PR #274, opened 2026-07-03
Fix duplicated push of node in HyperDHT initialization

zerotacg:
When running the cli and starting multiple nodes `--nodes 5` I'd expect it to start the number of nodes given

It actually counts every node twice so it never created the requested amount of nodes

```bash
> node bin.js --nodes 5
Booting DHT nodes...
Node #1 bound to { host: '0.0.0.0', family: 4, port: 46961 }
Node #3 bound to { host: '0.0.0.0', family: 4, port: 43422 }
Node #5 bound to { host: '0.0.0.0', family: 4, port: 52214 }
Fully started 5 Hyperswarm DHT nodes
```

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -53,8 +53,6 @@ async function startNodes(cnt, bootstrap) {
     const node = new HyperDHT({ host, port, anyPort: !port, bootstrap })
     await node.ready()
 
-    all.push(node)
-
     const id = all.push(node) - 1
     console.log('Node #' + id + ' bound to', node.address())
 


bin.js
```diff
@@ -53,8 +53,6 @@ async function startNodes(cnt, bootstrap) {
     const node = new HyperDHT({ host, port, anyPort: !port, bootstrap })
     await node.ready()
 
-    all.push(node)
-
     const id = all.push(node) - 1
```
zerotacg, line 56: As the id is only ever used for logging the ` - 1` could be removed to get the previous logging behaviour starting with `Node #1`.
Not sure if you wanted to express a zero based index for the node. So that change is more a matter of different tast.

lejeunerenard: Looks like here is were the double push came from and was zero based before:
https://github.com/holepunchto/hyperdht/pull/124/changes#diff-5fc985a6f001ff3e64da4eba221e879093a21518a38c19780dac89325ecd2c2fR58

lejeunerenard (approved):
Thank you! Great catch!

For posterity, the issue was introduced here #124 where a refactor on the `id` for a node accidentally left in the original `all.push(node)`.

PR #277, opened 2026-07-28
update address when flipping firewall

marco-hp:
This is a race condition

- Racer A: the node's own bootstrap self-check (dht-rpc's _bootstrap() → the "quick NAT heuristic" → _checkIfFirewalled()). This runs automatically once you start using a freshly-constructed HyperDHT, and it can flip the shared dht.firewalled boolean from true to false within milliseconds.

- Racer B: that node's own holepunch negotiation for the in-flight connection — specifically, the moment roundPunch() builds round 1's payload, it reads c.puncher.nat.firewall and c.puncher.nat.addresses at that instant and ships them to the peer. That per-connection Nat instance derives its firewall value from the same shared dht.firewalled boolean Racer A is mutating.

-- coworking with Claude

### DIFF
--- a/lib/nat.js
+++ b/lib/nat.js
@@ -140,6 +140,12 @@ module.exports = class Nat {
       return
     }
 
+    if (this.firewall === FIREWALL.OPEN) {
+      const addr = typeof this.dht.remoteAddress === 'function' ? this.dht.remoteAddress() : null
+      this.addresses = addr ? [{ host: addr.host, port: addr.port, hits: this.sampled || 1 }] : null
+      return
+    }
+
     if (this.firewall === FIREWALL.RANDOM) {
       this.addresses = [this._samplesHost[0]]
       return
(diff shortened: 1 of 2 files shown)


lejeunerenard:
@marco-hp Thanks for catching this! The `Nat` should return addresses even if the firewall is `OPEN`.

I pushed some changes as the solution doesn't account for samples that are available. I also adjusted the existing "zero-sample" test to use `.unfreeze()` as the path to update the addresses as `nat.update()` is not called outside of `add()`, though `.unfreeze()` itself is also never called, but theoretically could be. Finally I added a new test case to cover the case where there is a sample and the DHT flips to no longer being firewalled.

lib/nat.js
```diff
@@ -140,6 +140,17 @@ module.exports = class Nat {
       return
     }
 
+    if (this.firewall === FIREWALL.OPEN) {
+      if (this._samplesFull.length) {
+        this.addresses = [this._samplesFull[0]]
+        return
+      }
+
+      const addr = typeof this.dht.remoteAddress === 'function' ? this.dht.remoteAddress() : null
```
mafintosh: why needed? remoteAddress is always defined?

marco-hp: removed the check remoteAddress

PR #285, opened 2026-08-12
Add test coverage & clean up handshakes received when suspended & reorder clearing

lejeunerenard:
This PR came from explorations around potential memory leaks. It includes the following:

1. Added a test for guard against peer flooding a server with handshakes messages. Test called `peer cant flood w/ handshakes`
2. Reorder deleting `._connects.delete(k)` to before clearing holepunch entry.  
   Not required but is clearer that it always is cleared now and that a holepunch entry missing can't leak `_connects` though it was never observed to do so.
3. Add test & protect against leaking `_holepunches` entries when received during closing or being suspended. Test called `handshakes that arrive while the server is suspended are cleared`

`_clear deletes the _connects entry even when the slot was reused` credited to @HDegroote for writing.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -449,7 +449,10 @@ module.exports = class Server extends EventEmitter {
       }
     }
 
-    if (this._closing || this.suspended) return null
+    if (this._closing || this.suspended) {
+      this._clearLater(hs, id, k)
+      return null
+    }
 
     if (ourRemoteAddr || this._neverPunch) {
       hs.prepunching = setTimeout(onabort, HANDSHAKE_INITIAL_TIMEOUT)
--- a/lib/server.js
+++ b/lib/server.js
@@ -254,7 +254,10 @@ module.exports = class Server extends EventEmitter {
       return null
     }
 
-    if (this._closing || this.suspended) return null
+    if (this._closing || this.suspended) {
+      this._clearLater(hs, id, k)
+      return null
+    }
 
     try {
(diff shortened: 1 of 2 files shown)


mafintosh (dismissed):

lib/server.js
```diff
@@ -466,6 +474,7 @@ module.exports = class Server extends EventEmitter {
   }
 
   _clear(hs, id, k) {
+    this._connects.delete(k)
```
mafintosh: this doesnt look right, the check below is to make sure we are delting the right handshake

lejeunerenard: The reason for moving it up is to ensure that the handshake guard doesn't prevent the connection from being deleted. We have not observed that being the case and analyzing the code, I can't find how that'd happen. But it's also true in reverse, `_connects` is always updated with `_holepunches`.

The idea is that we always want to clear the `_connects` if we're trying to clear it's handshake regardless if we have it or not since the `k` uniquely represents the handshake.

lejeunerenard: Resolved out of band to revert (done here [c141992](https://github.com/holepunchto/hyperdht/pull/285/commits/c141992ccaf2a7aa36806c118712f4afa185e2d0)) as this move assumes a bug and would confuse the issue if it did exist, which we haven't recreated nor seems possible.

Better to assume correct behavior rather than be defensive for only a possibility of a bug and not actually addressing the bug.

PR #291, opened 2026-08-24
Destroy orphaned holepunchers when clearing handshakes

marcus-pousette-hp:
A firewalled server creates a Holepuncher (which binds a pool UDP socket) for every inbound handshake. When the punch is postponed with TRY_LATER (random punch limit) or the connection stays relayed, the prepunching timeout is cleared and nothing owns the puncher anymore. If the relayed raw stream then closes cleanly - a graceful client disconnect - the handshake record is cleared but the puncher and its socket survived until process restart, leaking one fd per such connection.

Destroy the puncher when its handshake record is cleared.

Co-authored with AI

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -315,10 +315,6 @@ module.exports = class Server extends EventEmitter {
       // (onsocket is called there only when the direct connection is established)
       const onrawstreamclose = () => {
         if (this._closing) return
-        if (hs.prepunching) {
-          clearTimeout(hs.prepunching)
-          hs.prepunching = null
-        }
         if (hs.puncher) {
           hs.puncher.onabort = noop
           hs.puncher.destroy()
--- a/lib/server.js
+++ b/lib/server.js
@@ -477,6 +477,13 @@ module.exports = class Server extends EventEmitter {
     if (id >= this._holepunches.length || this._holepunches[id] !== hs) return
     if (hs.clearing) clearTimeout(hs.clearing)
 
+    if (hs.puncher) {
+      // The handshake owns the puncher: destroy it here so a postponed or
+      // never-upgraded punch cannot leak its pool socket when cleared
+      hs.puncher.onabort = noop
+      hs.puncher.destroy()
+    }
+
     this._holepunches[id] = null
     while (
       this._holepunches.length > 0 &&
--- a/lib/server.js
+++ b/lib/server.js
@@ -315,6 +315,7 @@ module.exports = class Server extends EventEmitter {
       // (onsocket is called there only when the direct connection is established)
       const onrawstreamclose = () => {
         if (this._closing) return
+        if (hs.puncher) hs.puncher.destroy()
         this._clearLater(hs, id, k)
       }
       hs.rawStream.on('close', onrawstreamclose)
@@ -476,13 +477,6 @@ module.exports = class Server extends EventEmitter {
   _clear(hs, id, k) {
     if (id >= this._holepunches.length || this._holepunches[id] !== hs) return
     if (hs.clearing) clearTimeout(hs.clearing)
-
-    if (hs.puncher) {
-      // The handshake owns the puncher: destroy it here so a postponed or
-      // never-upgraded punch cannot leak its pool socket when cleared
-      hs.puncher.onabort = noop
-      hs.puncher.destroy()
-    }
 
     this._holepunches[id] = null
     while (
--- a/lib/server.js
+++ b/lib/server.js
@@ -315,7 +315,14 @@ module.exports = class Server extends EventEmitter {
       // (onsocket is called there only when the direct connection is established)
       const onrawstreamclose = () => {
         if (this._closing) return
-        if (hs.puncher) hs.puncher.destroy()
+        if (hs.prepunching) {
+          clearTimeout(hs.prepunching)
+          hs.prepunching = null
+        }
+        if (hs.puncher) {
+          hs.puncher.onabort = noop
+          hs.puncher.destroy()
+        }
         this._clearLater(hs, id, k)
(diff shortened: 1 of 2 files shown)


test/relaying.js
```diff
@@ -814,3 +814,87 @@ test.skip('server does not support connection relaying', async function (t) {
...
+  )
+  t.ok(!hs.puncher.destroyed, 'puncher still holds its socket')
+
+  // A graceful close can land before any relay error propagates, so the
+  // puncher cleanup must not depend on the raw stream error path
+  const punchSocket = hs.puncher.socket
+  hs.rawStream.destroy()
+
+  await once(punchSocket, 'close')
+  t.is(a._socketPool._sockets.size, 0, 'no orphaned holepunch sockets on the server')
```
lejeunerenard: Would be nice to assert that there are sockets before destroying the stream as well.

test/relaying.js
```diff
@@ -814,3 +814,98 @@ test.skip('server does not support connection relaying', async function (t) {
...
+  const Nat = require('../lib/nat')
+  const { FIREWALL } = require('../lib/constants')
+
+  const r = createDHT({ bootstrap, firewalled: false, ephemeral: true })
+  const a = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
+  const b = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
+
+  t.teardown(() => Promise.all([r.destroy(), a.destroy(), b.destroy()]), {
+    force: true,
+    order: 0
```
```suggestion
```

Its the default so not needed.

Why is the force needed here and throughout this test? The teardown for the Nat monkey patch makes sense to force but the rest are ephemeral resources for the test.

test/relaying.js
```diff
@@ -814,3 +814,98 @@ test.skip('server does not support connection relaying', async function (t) {
...
+  })
+  t.ok(hs, 'server parked a live puncher after relay pairing')
+  if (!hs) return
+
+  // Close the relayed raw stream without an error before direct upgrade. The
+  // puncher cleanup must not depend on the raw stream error path.
+  const punchSocket = hs.puncher.socket
+  hs.rawStream.destroy()
+
+  await waitFor(() => hs.puncher.destroyed && !a._socketPool.lookup(punchSocket))
```
lejeunerenard: If we keep `_clear()` as where it cleans up, instead of polling we should wait the timeout (`handshakeClearWait`) times a buffer percentage (maybe `handshakeClearWait * 1.2`?) to make it likely to be scheduled after. The waitFor assert the things we are checking so failing gives no info but a test timeout.

With destroying in the holepuncher, it seems it can be reduced to two setImmediate() promises, presumably one for letting the `close` event emit on the raw stream and another for the `close` to emit on the socket.

lib/server.js
```diff
@@ -477,6 +477,13 @@ module.exports = class Server extends EventEmitter {
     if (id >= this._holepunches.length || this._holepunches[id] !== hs) return
     if (hs.clearing) clearTimeout(hs.clearing)
 
+    if (hs.puncher) {
+      // The handshake owns the puncher: destroy it here so a postponed or
+      // never-upgraded punch cannot leak its pool socket when cleared
+      hs.puncher.onabort = noop
+      hs.puncher.destroy()
+    }
+
```
lejeunerenard: Shouldn't we destroy the puncher in `onrawstreamclose` instead? We dont need to keep the holepuncher around if the stream closed.

Elsewhere when we error we clean it up immediately rather than waiting on `_clear` to be called. For example `autoDestroy()` and `onabort()`.

test/relaying.js
marcus-pousette-hp: Good catch, got it working now with a pure event driven solution

lib/server.js
marcus-pousette-hp: ye that is way cleaner! good catch :) Fixed!

test/relaying.js
marcus-pousette-hp: Fixed. The force was a remnant when I was debugging an unexpected test error. Thanks

lib/server.js
```diff
@@ -315,6 +315,7 @@ module.exports = class Server extends EventEmitter {
       // (onsocket is called there only when the direct connection is established)
       const onrawstreamclose = () => {
         if (this._closing) return
+        if (hs.puncher) hs.puncher.destroy()
```
lejeunerenard: Do we not need to set `onabort` to `null`? Prepunching is take care of by the relay connection & setting `aborted` is only used by the relay stream closing the rawStream.

test/relaying.js
```diff
@@ -814,3 +814,122 @@ test.skip('server does not support connection relaying', async function (t) {
...
+
+  await Promise.all([r.fullyBootstrapped(), a.fullyBootstrapped(), b.fullyBootstrapped()])
+
+  const relay = new RelayServer({
+    createStream(opts) {
+      return r.createRawStream({ ...opts, framed: true })
+    }
+  })
+
+  t.teardown(() => relay.close(), { order: -10 })
```
lejeunerenard: nit: Do we need the `order` settings in this test?

lib/server.js
marcus-pousette-hp: You are right. Failed to cleanup properly after moving from the first implementation that destroyed puncher inside _clear. Fixed!

test/relaying.js
marcus-pousette-hp: No. And especially wrong without force flags, FIxed!

lib/server.js
marcus-pousette-hp: I disabled it with noop now

lejeunerenard (approved):

PR #292, opened 2026-08-24
Fix/clear handshake on relay abort

marcus-pousette-hp:
## Leaks (both reproducible on `main`)

- **Relay pairing fails while the punch is parked.** Nothing re-arms a timer after `TRY_LATER`. If the client stops retrying, the unconnected raw stream and puncher can remain retained indefinitely without another cleanup event. This server-side retention path remains after #291.
- **Relay dies after pairing.** The server stopped watching the relay once paired. Without direct recovery or another stream-close event, the connection can remain stuck; keepalive can eventually clean it up if enabled.

## Fix

- Listen for `close` on the blind-relay Client, so relay loss is seen before and after pairing.
- On relay loss: if the punch already aborted, tear down now. Otherwise wait `relayRecoveryWait` (30 s, anchored to the failure) for the direct path, then tear down unless a round, a punch or the prepunch timer is still active.
- Teardown reuses the existing raw-stream close path. New server option `relayRecoveryWait`.

## Relay lifecycle

These diagrams zoom in on the server-side blind-relay lifecycle within holepunching; NAT negotiation and individual punch rounds are omitted. `TRY_LATER` parks the direct punch, not the relay: it can happen during setup or after pairing, and a later round may resume the direct attempt.

**Before (`main`).** Pairing failure adds no cleanup deadline for a parked punch; after pairing, relay closure is not observed. The dashed leak path applies only if direct recovery and other cleanup do not happen.

```mermaid
%%{init: {"flowchart": {"curve": "linear", "nodeSpacing": 30, "rankSpacing": 45}}}%%
flowchart TB
  S["Relay setup<br/>punch may be parked<br/>on TRY_LATER"]
  R(["Relayed only<br/>punch may be parked<br/>on TRY_LATER"])
  U[Relay unavailable]
  D([Directly connected])
  F([Connection failed])
  L(["Resources retained /<br/>leak"])

  S -->|paired| R
  S -->|pairing fails| U
  R -.->|relay closes; not observed| U
  S -->|direct route wins| D
  R -->|direct upgrade| D
  U -->|direct recovery| D
  U -->|existing cleanup event| F
  U -.->|still parked / stalled; no cleanup| L
```

**After (this PR).** Relay closure is observed after pairing too. A parked `TRY_LATER` punch gets the same direct-recovery grace as other not-yet-aborted attempts; parking alone does not trigger immediate cleanup. Bold arrows show the new handling.

```mermaid
%%{init: {"flowchart": {"curve": "linear", "nodeSpacing": 30, "rankSpacing": 45}}}%%
flowchart TB
  S["Relay setup<br/>punch may be parked<br/>on TRY_LATER"]
  R(["Relayed only<br/>punch may be parked<br/>on TRY_LATER"])
  U[Relay unavailable]
  G["Direct recovery<br/>30 s grace by default"]
  D([Directly connected])
  F([Connection failed])

  S -->|paired| R
  S -->|pairing fails| U
  R ==>|relay closes; observed| U
  S -->|direct route wins| D
  R -->|direct upgrade| D
  U ==>|already aborted| F
  U ==>|not aborted| G
  G -->|direct route wins| D
  G ==>|grace elapsed; no active work| F
```

- **Recovery grace:** `relayRecoveryWait` starts at relay failure, not at `TRY_LATER`. Direct recovery can win during the grace period or while protected work continues afterward. An in-flight round, started punch or prepunch timer defers cleanup; 30 s is not a hard lifetime limit. A completing round rechecks cleanup, while punches and prepunch timers retain their existing success/abort paths. Later probes do not extend the deadline, and a healthy relay never starts it.
- **Reading the outcomes:** “Relayed only” carries traffic but may still upgrade or lose its relay. “Relay unavailable” describes the transport, not proof the old server detected its loss. The before graph can still clean up through a subsequent stream-close event, including application close or keepalive.

## Why this shape

- **30 s grace:** relay loss must not abort a direct connection that can still complete; pairing times out inside the client's 10-20 s backoff.
- **In-flight counter:** a deadline landing mid-round lets the round finish.
- **Anchored deadline:** later rounds cannot extend it.
- **Started by relay loss, not by silence:** a holepunch-side idle timer is smaller and passes the same tests, but a client waiting on its own random-punch budget sends one probe and then stays silent 20-55 s per queued punch. An idle timer kills the parked puncher on a healthy relayed connection, and the connection never upgrades. A clock that starts only at relay loss cannot touch a healthy connection.

Each point has a test that fails when the piece is removed.

## Tests

- paired relay loss eventually clears a parked TRY_LATER handshake
- relay transport failure while pairing clears a parked handshake
- relay transport failure after pairing clears an unrecoverable handshake
- non-punching probes do not extend the relay recovery deadline
- two relay-failure cases in "relay connection upgrades to direct connection"
- relayed connection still upgrades when the client pauses between rounds

The first three fail on `main`. The last passes on `main` and here, and fails on an idle-timer variant.

Follow-up, not here: the client-side mirror in `lib/connect.js` (no `close` listener after pairing, no `ABORTED` round when it gives up).

Co-authored with AI

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -315,10 +315,6 @@ module.exports = class Server extends EventEmitter {
       // (onsocket is called there only when the direct connection is established)
       const onrawstreamclose = () => {
         if (this._closing) return
-        if (hs.prepunching) {
-          clearTimeout(hs.prepunching)
-          hs.prepunching = null
-        }
         if (hs.puncher) {
           hs.puncher.onabort = noop
           hs.puncher.destroy()
--- a/lib/server.js
+++ b/lib/server.js
@@ -477,6 +477,13 @@ module.exports = class Server extends EventEmitter {
     if (id >= this._holepunches.length || this._holepunches[id] !== hs) return
     if (hs.clearing) clearTimeout(hs.clearing)
 
+    if (hs.puncher) {
+      // The handshake owns the puncher: destroy it here so a postponed or
+      // never-upgraded punch cannot leak its pool socket when cleared
+      hs.puncher.onabort = noop
+      hs.puncher.destroy()
+    }
+
     this._holepunches[id] = null
     while (
       this._holepunches.length > 0 &&
--- a/lib/server.js
+++ b/lib/server.js
@@ -19,6 +19,7 @@ const { ALREADY_LISTENING, NODE_DESTROYED, KEYPAIR_ALREADY_USED } = require('./e
 
 const HANDSHAKE_CLEAR_WAIT = 10000
 const HANDSHAKE_INITIAL_TIMEOUT = 10000
+const RELAY_RECOVERY_WAIT = 30000
 
 module.exports = class Server extends EventEmitter {
   constructor(dht, opts = {}) {
@@ -32,6 +33,7 @@ module.exports = class Server extends EventEmitter {
     this.holepunch = opts.holepunch || (() => true)
     this.relayThrough = opts.relayThrough || null
     this.relayKeepAlive = opts.relayKeepAlive || 5000
+    this.relayRecoveryWait = opts.relayRecoveryWait || RELAY_RECOVERY_WAIT
     this.pool = opts.pool || null
     this.createHandshake = opts.createHandshake || defaultCreateHandshake
     this.createSecretStream = opts.createSecretStream || defaultCreateSecretStream
@@ -141,6 +143,7 @@ module.exports = class Server extends EventEmitter {
       if (h && h.puncher) h.puncher.destroy()
       if (h && h.clearing) clearTimeout(h.clearing)
       if (h && h.prepunching) clearTimeout(h.prepunching)
+      if (h) clearRelayRecovery(h)
       if (h && h.rawStream) h.rawStream.destroy()
     }
 
@@ -239,7 +242,8 @@ module.exports = class Server extends EventEmitter {
       relaySocket: null,
       relayClient: null,
       relayPaired: false,
-      relayFailed: false,
+      relayFailedAt: null,
+      relayRecoveryTimeout: null,
       validUpgrade: true
     }
 
@@ -316,6 +320,7 @@ module.exports = class Server extends EventEmitter {
       // This can happen on a relayed connection which never connects directly
       // (onsocket is called there only when the direct connection is established)
       const onrawstreamclose = () => {
+        clearRelayRecovery(hs)
         if (this._closing) return
         if (hs.puncher) {
           hs.puncher.onabort = noop
@@ -328,6 +333,7 @@ module.exports = class Server extends EventEmitter {
       hs.onsocket = (socket, port, host) => {
         if (hs.rawStream === null) return // Already hole punched
 
+        clearRelayRecovery(hs)
         this._clearLater(hs, id, k)
 
         if (hs.prepunching) {
@@ -424,6 +430,7 @@ module.exports = class Server extends EventEmitter {
 
     const onabort = () => {
       hs.aborted = true
+      clearRelayRecovery(hs)
       if (hs.prepunching) clearTimeout(hs.prepunching)
       hs.prepunching = null
       if (hs.puncher) {
@@ -480,6 +487,7 @@ module.exports = class Server extends EventEmitter {
   }
 
   _clear(hs, id, k) {
+    clearRelayRecovery(hs)
     if (id >= this._holepunches.length || this._holepunches[id] !== hs) return
     if (hs.clearing) clearTimeout(hs.clearing)
 
@@ -638,7 +646,7 @@ module.exports = class Server extends EventEmitter {
       }
     } finally {
       h.activeHolepunchRequests--
-      maybeDestroyRelayHandshake(h)
+      maybeDestroyRelayHandshake(h, this.relayRecoveryWait)
     }
   }
 
@@ -667,6 +675,7 @@ module.exports = class Server extends EventEmitter {
 
   _relayConnection(hs, relayThrough, remotePayload, h) {
     this.dht.stats.relaying.attempts++
+    const relayRecoveryWait = this.relayRecoveryWait
 
     let isInitiator
     let publicKey
@@ -727,29 +736,50 @@ module.exports = class Server extends EventEmitter {
     function onabort() {
       if (hs.relaySocket !== relaySocket || hs.relayClient !== relayClient) return
       if (!hs.relayPaired) dht.stats.relaying.aborts++
-      hs.relayFailed = true
+      if (hs.relayFailedAt === null) hs.relayFailedAt = Date.now()
       destroyRelayConnection(hs)
-      maybeDestroyRelayHandshake(hs)
+      maybeDestroyRelayHandshake(hs, relayRecoveryWait)
     }
   }
 }
 
-function maybeDestroyRelayHandshake(hs) {
-  if (!hs.relayFailed || hs.rawStream === null || hs.rawStream.destroyed) return
+function maybeDestroyRelayHandshake(hs, relayRecoveryWait) {
+  if (hs.relayFailedAt === null || hs.rawStream === null || hs.rawStream.destroyed) return
 
   if (hs.aborted) {
     hs.rawStream.destroy()
     return
   }
 
-  // A pending prepunch or active holepunch workflow can still recover by
-  // upgrading the raw stream. Recheck when the active workflow completes.
+  // Holepunching spans multiple requests. Give the next round time to arrive,
+  // but anchor the deadline to relay failure so requests cannot extend it.
+  const remaining = relayRecoveryWait - (Date.now() - hs.relayFailedAt)
+  if (remaining > 0) {
+    if (hs.relayRecoveryTimeout !== null) return
+
+    const rawStream = hs.rawStream
+    hs.relayRecoveryTimeout = setTimeout(() => {
+      hs.relayRecoveryTimeout = null
+      if (hs.rawStream !== rawStream) return
+      maybeDestroyRelayHandshake(hs, relayRecoveryWait)
+    }, remaining)
+    hs.relayRecoveryTimeout.unref()
+    return
+  }
+
+  // An active holepunch workflow owns cleanup after the recovery deadline.
   if (hs.prepunching !== null || hs.activeHolepunchRequests > 0) return
 
   const p = hs.puncher
   if (p && p.punching) return
 
   hs.rawStream.destroy()
+}
+
+function clearRelayRecovery(hs) {
+  if (hs.relayRecoveryTimeout !== null) clearTimeout(hs.relayRecoveryTimeout)
+  hs.relayRecoveryTimeout = null
+  hs.relayFailedAt = null
 }
 
 function isConsistent(fw) {
--- a/lib/server.js
+++ b/lib/server.js
@@ -315,6 +315,7 @@ module.exports = class Server extends EventEmitter {
       // (onsocket is called there only when the direct connection is established)
       const onrawstreamclose = () => {
         if (this._closing) return
+        if (hs.puncher) hs.puncher.destroy()
         this._clearLater(hs, id, k)
       }
       hs.rawStream.on('close', onrawstreamclose)
@@ -476,13 +477,6 @@ module.exports = class Server extends EventEmitter {
   _clear(hs, id, k) {
     if (id >= this._holepunches.length || this._holepunches[id] !== hs) return
     if (hs.clearing) clearTimeout(hs.clearing)
-
-    if (hs.puncher) {
-      // The handshake owns the puncher: destroy it here so a postponed or
-      // never-upgraded punch cannot leak its pool socket when cleared
-      hs.puncher.onabort = noop
-      hs.puncher.destroy()
-    }
 
     this._holepunches[id] = null
     while (
--- a/lib/server.js
+++ b/lib/server.js
@@ -699,7 +699,6 @@ module.exports = class Server extends EventEmitter {
     hs.relayClient = relayClient
     relaySocket.setKeepAlive(this.relayKeepAlive)
     hs.relayTimeout = setTimeout(onabort, 15000)
-    relaySocket.once('close', onabort)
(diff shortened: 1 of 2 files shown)


lib/server.js
```diff
@@ -709,7 +716,21 @@ module.exports = class Server extends EventEmitter {
...
+        hs.rawStream.destroy()
+        return
+      }
+
+      // Without pairing the raw stream never connects and a parked puncher
+      // has no timer, so nothing would ever clear this handshake. Destroy the
+      // raw stream so the clearing gets scheduled. Punchers that are punching
+      // or connected clean up through their own paths.
+      const p = hs.puncher
+      if (!hs.relayPaired && hs.prepunching === null && (!p || (!p.punching && !p.connected))) {
```
lejeunerenard:
```suggestion
      if (!hs.relayPaired && hs.prepunching === null && (!p || p.punching || p.connected)) {
```

Easier to understand in the positive and since its equivilant (De Morgan's law) I think this will be easier to read.

marcus-pousette-hp: Fixed

test/relaying.js
```diff
@@ -814,3 +818,207 @@ test.skip('server does not support connection relaying', async function (t) {
   await b.destroy()
   await c.destroy()
 })
+
+test('relayed TRY_LATER stream close destroys the parked server puncher', async function (t) {
```
lejeunerenard: Left over from when the PR was stacked.

lib/server.js
```diff
@@ -709,7 +716,23 @@ module.exports = class Server extends EventEmitter {
...
+      if (hs.aborted) {
+        hs.rawStream.destroy()
+        return
+      }
+
+      // Without pairing the raw stream never connects and a parked puncher
+      // has no timer, so nothing would ever clear this handshake. Destroy the
+      // raw stream so the clearing gets scheduled. Punchers that are punching
+      // or connected clean up through their own paths.
+      if (hs.relayPaired || hs.prepunching !== null) return
```
lejeunerenard: Should it not potentially destroy the stream if `hs.relayPaired` is true? That would mean this came from an error on the client since it will have had to succeed via the `data` event.

This does point out that there is probably another edgecase here as the `error` event is on the blind relay request, not on the client. So that `error` event wouldn't ever happen with the `hs.relayPaired` as true from that event. We likely aren't covering the `error` event on the relay stream as well.

Regardless, if the relay pairing errors and closes and the raw stream cant be recovered by the holepunching, then shouldn't the stream be destroyed?

lejeunerenard: Example test for case (generated by AI) where even without a holepuncher the relay stream can leak.

```js
test('relay client error after successful pairing clears the handshake', async function (t) {

  const relayNode = createDHT({ bootstrap })
  const serverNode = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })
  const clientNode = createDHT({ bootstrap, quickFirewall: false, ephemeral: true })

  t.teardown(() => Promise.all([relayNode.destroy(), serverNode.destroy(), clientNode.destroy()]))

  // Report a remote address so `_addHandshake` takes the ourRemoteAddr fast
  // path: no Holepuncher is ever created, so relay pairing is the only thing
  // that can ever complete or clear this handshake.
  serverNode.remoteAddress = () => ({ host: '1.2.3.4', port: 9999 })

  const relayServer = new RelayServer({
    createStream(opts) {
      return relayNode.createRawStream({ ...opts, framed: true })
    }
  })

  t.teardown(() => relayServer.close())

  const relayTransportServer = relayNode.createServer(function (socket) {
    relayServer.accept(socket, { id: socket.remotePublicKey }).on('error', () => {})
  })

  await relayTransportServer.listen()

  const appServer = serverNode.createServer(
    {
      relayThrough: relayTransportServer.publicKey,
      shareLocalAddress: false,
      handshakeClearWait: 100
    },
    function (socket) {
      socket.on('error', () => {})
    }
  )

  await appServer.listen()

  const clientSocket = clientNode.connect(appServer.publicKey, {
    fastOpen: false,
    localConnection: false,
    holepunch: () => false
  })
  clientSocket.on('error', () => {})
  t.teardown(() => clientSocket.destroy())

  await new Promise((resolve, reject) => {
    clientSocket.once('open', resolve)
    clientSocket.once('error', reject)
  })

  const hs = await waitFor(() => appServer._holepunches.find((h) => h && h.relayPaired))
  t.absent(hs.puncher, 'no puncher was ever needed to complete this handshake')
  t.ok(hs.rawStream && !hs.rawStream.destroyed, 'server rawStream is alive right after pairing')

  const rawStream = hs.rawStream

  // Simulate the relay dying after pairing already succeeded: nothing but a
  // listener on hs.relaySocket can notice this, since the one-shot pair()
  // request already completed and detached itself.
  hs.relaySocket.destroy(new Error('simulated relay death'))

  await waitFor(() => rawStream.destroyed, 2000)
  t.ok(rawStream.destroyed, 'server destroyed the orphaned rawStream promptly')

  await waitFor(() => !appServer._holepunches.includes(hs), 2000)
  t.absent(
    appServer._holepunches.find((h) => h === hs),
    'server cleared the handshake after the relay died'
  )
})
```

marcus-pousette-hp: I forgot to answer here, but this has been implemented and fixed

marcus-pousette-hp: The stream is not destroyed the moment the relay dies, because the direct holepunch may still be in flight and can still succeed. The recovery time in this PR gives it time to win before the server gives up. And the grace period is used to fixed the leak anyway here

test/relaying.js
```diff
@@ -545,7 +580,20 @@ test('relay connection upgrades to direct connection', async function (t) {
...
+      relayFailure: true
+    },
+    {
+      name: 'server relay loss past deadline during a punching request',
+      connectionKeepAlive: false,
+      relayFailure: true,
+      remoteHolepunching: true,
+      relayRecoveryWait: 100,
+      waitForRecoveryExpiry: true
+    }
```
marcus-pousette-hp, line 596: We have been previously favored split tests with helper functions instead of arrays of cases in one test, please let me know if this is preferred. Personally in this scenario I like this approach because the cases are very similiar

lib/server.js
```diff
@@ -716,13 +738,52 @@ module.exports = class Server extends EventEmitter {
...
+    hs.relayRecoveryTimeout = setTimeout(() => {
+      hs.relayRecoveryTimeout = null
+      maybeDestroyRelayHandshake(hs, relayRecoveryWait)
+    }, remaining)
+    hs.relayRecoveryTimeout.unref()
+    return
+  }
+
+  // An active holepunch workflow owns cleanup after the recovery deadline.
+  if (hs.prepunching !== null || hs.activeHolepunchRequests > 0) return
```
lejeunerenard, line 773: This should be above setting up the timer right? Why wait for holepunching if it's already inactive and can't own the recovery.

lib/server.js
```diff
@@ -716,13 +738,52 @@ module.exports = class Server extends EventEmitter {
 
     const dht = this.dht
     function onabort() {
+      if (hs.relaySocket !== relaySocket || hs.relayClient !== relayClient) return
```
lejeunerenard, line 741: Can the relaySocket or client be swapped to be different? Aka why not a null check.

And why do we need to test both? Shouldn't they be destroyed together? Or at least if we have a socket then we should have a client right?

lib/server.js
```diff
@@ -716,13 +738,52 @@ module.exports = class Server extends EventEmitter {
...
       if (!hs.relayPaired) dht.stats.relaying.aborts++
```
lejeunerenard, line 742: This now includes the `close` event of the relay client, I started looking at a potential double triggering scenario (which is accounted for via the guard above) and realized that `onabort` triggers for normal closure too.

For example, it will be `1` for node `a` (the server) at the end of the test `relay connections through node, client side` with a brief 100ms delay (to let the messages propagate) before destroying the nodes. So it ends not in an error state but still marks the relaying as aborted. IE update the ending of the test to this will fail:
```js
  aSocket
    // ...
    .end('hello world')
  await lc

  await new Promise((resolve) => setTimeout(resolve, 100))
  t.is(a.stats.relaying.aborts, 0, 'A never aborts')

  await a.destroy()
```

lib/server.js
marcus-pousette-hp: Inactive at relay loss usually just means between rounds: the client comes back within its 10-20 s backoff and the punch can still win. Checking first would effectively be a zero grace period

lib/server.js
marcus-pousette-hp: You are right, they are always set and cleared together so this can be simplified (fixed)

lib/server.js
marcus-pousette-hp: Good catch! Fixed.

marcus-pousette-hp:
The diagrams here deliberately focus on the relay lifecycle changed by this PR. I did an attempt to do the full state machine including the TRY_LATER path and all the different states but it because too big to include in the PR. I will post one graph below here,  what I got so far if this PR would be included. Perhaps opening a PR adding it separately once (if) this PR is approved to the documentation.


```mermaid
%%{init: {"flowchart": {"curve": "linear", "nodeSpacing": 25, "rankSpacing": 45}}}%%
flowchart TB
  S[Handshake accepted]

  subgraph relaying [Relay being set up]
    A1["Negotiating /<br/>punching"]
    A2["TRY_LATER /<br/>parked"]
    A3[Holepunch aborted]

    A1 -->|TRY_LATER| A2
    A2 -->|resume round| A1
    A1 -->|abort or timeout| A3
    A2 -->|client abort| A3
  end

  subgraph relayed [Relay paired]
    B1["Negotiating /<br/>punching"]
    B2["TRY_LATER /<br/>parked"]

    B1 -->|TRY_LATER| B2
    B2 -->|resume round| B1
  end

  subgraph norelay [No usable relay]
    C1["Negotiating /<br/>punching"]
    C2["Waiting for<br/>direct recovery"]
    Q{Already aborted?}

    C1 -->|relay failed; between rounds| C2
    C2 -->|resume round| C1
  end

  D([Directly connected])
  R([Relayed only])
  F([Connection failed])

  S -->|relay configured| A1
  S -->|no relay| C1
  S -->|open or LAN shortcut| D

  A1 -->|paired| B1
  A2 -->|paired| B2
  A3 -->|paired| R
  A1 -->|paired; no puncher needed| R

  A1 -->|pairing fails| C1
  A2 -->|pairing fails| C2
  A3 -->|pairing fails| F

  B1 -->|abort or timeout| R
  B2 -->|client abort| R
  C1 -->|abort, timeout or budget denied| F
  C2 -->|client abort| F

  A1 -->|direct route wins| D
  A2 -->|direct route wins| D
  B1 -->|direct route wins| D
  B2 -->|direct route wins| D
  C1 -->|direct route wins| D
  C2 -->|direct route wins| D

  B1 ==>|relay lost| C1
  B2 ==>|relay lost| C2
  R ==>|relay lost| Q
  Q ==>|yes| F
  Q ==>|no| C2

  C1 ==>|deadline reached; work still active| C1
  C2 ==>|deadline reached; no active work| F
```

lib/server.js
lejeunerenard: Yeah i omitted the context that this question presumes that the holepuncher / direct path part would own whether to clean itself up given it would be mid punching.

test/relaying.js
```diff
@@ -213,6 +228,25 @@ async function waitFor(fn, timeout = 2000) {
...
+    server._clear = function (...args) {
+      const result = clear.apply(this, args)
+      if (args[0] === hs) {
+        server._clear = clear
+        resolve()
+      }
+      return result
+    }
+  })
+}
```
lejeunerenard: nit: We should move these to the bottom. Not a blocker as we need to move `waitFor` as well. Probably just a maintenance PR for moving things around for consistent style.

test/relaying.js
```diff
@@ -639,7 +696,30 @@ test('relay connection upgrades to direct connection', async function (t) {
 
     const clientUpgraded = once(clientSocket.rawStream, 'remote-changed')
     const serverUpgraded = once(serverSocket.rawStream, 'remote-changed')
-    const relaySocketsClosed = relaySockets.map((socket) => once(socket, 'close'))
+    const relaySocketsClosed = relaySockets.map((socket) =>
+      opts.relayFailure ? closed(socket) : once(socket, 'close')
```
```suggestion
      closed(socket)
```

Don't know why we'd need both types of promisified `close` events. Either case if there is errors we want to know about them.

lib/server.js
marcus-pousette-hp: Aaah I get it now. Good catch. This would reduce some unecessary timers. Fixed!  https://github.com/holepunchto/hyperdht/pull/292/commits/048429341a22c403846c7c3c9500a39c34d5aa8e

lejeunerenard: Thanks! I think that's cleaner.

Actually, yielding to the puncher / direct path here means that if the holepuncher has active requests and/or still considers itself punching and the timer goes off, then it will still rely on the puncher to clean up the handshake / rawStream anyways. So the relay recovery timeout isn't an upper bound on waiting and we rely on the holepuncher / direct path.

The timer does limit the parked holepuncher, so more just evidence that we might want to refactor so the timer / state is on the holepuncher / direct path side.

lib/server.js
```diff
@@ -716,13 +738,53 @@ module.exports = class Server extends EventEmitter {
...
+
+  // An active holepunch workflow owns cleanup.
+  if (hs.prepunching !== null || hs.activeHolepunchRequests > 0) return
+
+  const p = hs.puncher
+  if (p && p.punching) return
+
+  // Holepunching spans multiple requests. Give the next round time to arrive,
+  // but anchor the deadline to relay failure so requests cannot extend it.
+  const remaining = relayRecoveryWait - (Date.now() - hs.relayFailedAt)
```
lejeunerenard: If no holepuncher is created (such as [here](https://github.com/holepunchto/hyperdht/blob/048429341a22c403846c7c3c9500a39c34d5aa8e/lib/server.js#L470-L473)), we keep the connection for the full timeout since none of the guards above will trigger.

We should add a check like the following above setting up the timer:
```js
if (!p) {
  hs.rawStream.destroy()
  return
}
```

PR #293, opened 2026-08-26
expose stats socket pool

marco-hp:


lib/socket-pool.js
```diff
@@ -18,11 +23,13 @@ module.exports = class SocketPool {
 
   _add(ref) {
     this._sockets.set(ref.socket, ref)
+    this.stats.socketsAdded++
   }
 
   _remove(ref) {
     this._sockets.delete(ref.socket)
     this._lingering.delete(ref)
+    this.stats.socketsRemoved++
```
marcus-pousette-hp: when bind fails we need to do this to prevent accounting errors (formatting is not good here when doing Github suggestion manually)

```suggestion
const removed = this._sockets.delete(ref.socket)
this._lingering.delete(ref)
if(removed)  this.stats.socketsRemoved++
```

marcus-pousette-hp (changes requested):

lib/socket-pool.js
marco-hp: updated

lib/server.js
```diff
@@ -585,6 +585,7 @@ module.exports = class Server extends EventEmitter {
           this.dht._randomPunches >= this.dht._randomPunchLimit ||
           Date.now() - this.dht._lastRandomPunch < this.dht._randomPunchInterval
         ) {
+          this.dht.stats.punches.tryLater++
```
marcus-pousette-hp, line 588: For #291 and #292 diagnostics it is more meaningful if this is once per handshake. Now it increments on each TRY_LATER reply. 

This increment should be below the 
```if(!h.relayToken) return ```

File: test/stats.js [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+    relaying: { attempts: 0, successes: 0, aborts: 0 },
+    queries: { active: 0, total: 0 },
+    requests: { active: 0, total: 0, responses: 0, timeouts: 0, retries: 0 },
+    commands: {
+      ping: { tx: 0, rx: 0 },
+      pingNat: { tx: 0, rx: 0 },
+      findNode: { tx: 0, rx: 0 },
+      downHint: { tx: 0, rx: 0 }
+    },
+    socketPool: { socketsAdded: 0, socketsRemoved: 0 }
```
marcus-pousette-hp:
```ts
queries: {
  active,
  total
}

requests: {
  active,
  total,
  responses,
  timeouts,
  retries
}

commands: {
  ping: { tx, rx },
  pingNat: { tx, rx },
  findNode: { tx, rx },
  downHint: { tx, rx }
}
```

these seems to be owned by dht-rpc so if there is a future addition of stats in dht-rpc this test would break in in this repo

lib/server.js
marcus-pousette-hp: so something like 


```ts
if (!h.relayToken) return this._abort(h, ERROR.TRY_LATER)

if (!h.tryLaterCounted) {
  h.tryLaterCounted = true
  this.dht.stats.punches.tryLater++
}
```


would be ideal

marcus-pousette-hp (changes requested):

lib/server.js
lejeunerenard: Incrementing once per TRY_LATER is more correct in general since both paths are erroring. It does make it vaguer for debugging #291 & #293, but these stats will be used for more than that case. Also it is possible for a peer to be told to try later twice if another peer starts a holepunch with the server in between.

test/stats.js
marco-hp: updated

lib/server.js
marco-hp: @lejeunerenard @marcus-pousette-hp can we confirm the current code is ok or needs to update?

or shall we add another stat for tryLater for relayToken condition?

marcus-pousette-hp: @lejeunerenard what if we do both?

(another stat)

```ts
this.dht.stats.punches.tryLater++

if (!h.relayToken) return this._abort(h, ERROR.TRY_LATER)

if (!h.tryLaterCounted) {
  h.tryLaterCounted = true
  this.dht.stats.punches.tryLaterRelayedHandshakes++
}
```

lejeunerenard: Adding a specific stat makes sense since we can't correlate two stats together. I dont think the gate is necessary though as it can be triggered multiple times which we'd want to know in general.

marco-hp: updated

marcus-pousette-hp (approved):

lejeunerenard (approved):

